PR #38, opened 2015-01-15
Update to use buildpack-deps:*-scm as the base

### LINKED ISSUE #37: Include SSH for private git repos
It seems the golang images don't have ssh installed. This prevents using git to pull things from ssh repos not hosted with a git daemon.

Could ssh be added to the image?

I just found this issue: https://github.com/docker-library/golang/issues/33 which has valid points. My use case is using the golang image as a base in ci/cd system (drone), so I need to be able to pull from private repos.

tianon:
Fixes #37

### DIFF
--- a/1.3/Dockerfile
+++ b/1.3/Dockerfile
@@ -1,9 +1,8 @@
-FROM debian:jessie
+FROM buildpack-deps:jessie-scm
 
-# SCMs for "go get", gcc for cgo
+# gcc for cgo
 RUN apt-get update && apt-get install -y \
-		ca-certificates curl gcc libc6-dev make \
-		bzr git mercurial \
+		gcc libc6-dev make \
 		--no-install-recommends \
 	&& rm -rf /var/lib/apt/lists/*
 
--- a/1.3/wheezy/Dockerfile
+++ b/1.3/wheezy/Dockerfile
@@ -1,9 +1,8 @@
-FROM debian:wheezy
+FROM buildpack-deps:wheezy-scm
 
-# SCMs for "go get", gcc for cgo
+# gcc for cgo
 RUN apt-get update && apt-get install -y \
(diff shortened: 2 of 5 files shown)


yosifkit:
LGTM

tianon:
cc @proppy

1.3/Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM debian:jessie
+FROM buildpack-deps:jessie-micro
 
-# SCMs for "go get", gcc for cgo
+# gcc for cgo
 RUN apt-get update && apt-get install -y \
-		ca-certificates curl gcc libc6-dev make \
-		bzr git mercurial \
+		gcc libc6-dev make \
```
proppy: is ca-certificates in jessie or buildpacks-micro?

proppy:
maybe we could add some docs on why we depends on buildpack-deps and what it includes.

tianon:
Ok.  I've updated this to use the new (and more specific / well-defined) `scm` variant of `buildpack-deps`, which is much more exact about what is included.  See https://registry.hub.docker.com/_/buildpack-deps/ for more information, especially about which packages we've got.

tianon:
(under "What's included?")

proppy:
Nice, LGTM. Out of curiosity is there a specific reason to have "curl" && "wget" in the -curl variant?

1.3/Dockerfile
proppy: I think you need ca-certificates too.

tianon:
Only that the size difference is reasonably close to negligible and then either is usable depending on personal user taste -- "curl" just seemed like an appropriate general name for "common HTTP(s) fetching tools" (like "scm" has all the common Source Control Management tools).

PR #41, opened 2015-02-05
Remove set -x to prevent spamming stderr

cortex:
In our environment, we assume that everything coming from a docker app written to stderr is an error message. The set -x in the docker base image means we get a spurious "+ exec app" message on app startup.

### DIFF
--- a/go-wrapper
+++ b/go-wrapper
@@ -74,17 +74,17 @@ case "$cmd" in
 	download)
 		execCommand=( go get -v -d "$@" )
 		if [ "$goDir" ]; then execCommand+=( "$goDir" ); fi
-		set -x; exec "${execCommand[@]}"
+		exec "${execCommand[@]}"
 		;;
 		
 	install)
 		execCommand=( go install -v "$@" )
 		if [ "$goDir" ]; then execCommand+=( "$goDir" ); fi
-		set -x; exec "${execCommand[@]}"
+		exec "${execCommand[@]}"
 		;;
 		
 	run)
-		set -x; exec "$goBin" "$@"
+		exec "$goBin" "$@"
 		;;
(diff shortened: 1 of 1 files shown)


tianon:
I'm personally -1 on this change, especially since this can be trivially avoided by using one of the following patterns:
```Dockerfile
FROM golang:1.4-onbuild
CMD ["app"]
```
or: (assuming use of `// import ".../myapp"` comments for "canonical import path" support)
```Dockerfile
FROM golang:1.4-onbuild
CMD ["myapp"]
```

cortex:
Even if it is trivial to get rid of, I don't really see the usefulness of keeping it. Thanks for the simple workarounds though! IMO, it should at least be documented.

tianon:
I definitely fully agree that it should be documented, but I disagree that
it isn't useful.  Without it, there's no indication of which binary was
actually executed.

cortex:
I don't think that will be very useful for the end-user in most cases, since the binary name is generated by these ONBUILD scripts anyway. I prefer my apps to be silent unless they have something useful to tell me. My particular itch is solved with the workaround you gave me, but for the record two other approaches that could make life easier for future users:
1) add a flag to the script for debugging this
2) echo $goBin to stdout instead

cortex:
Here is some evidence that I'm not the only one that having a problem with this (plus an uglier workaround).:
https://registry.hub.docker.com/u/bjaglin/registree/dockerfile/

tianon:
https://github.com/search?q=from+golang%3Aonbuild&type=Code&utf8=%E2%9C%93 seems to suggest that most people aren't bothered by it terribly (or they use `CMD` or `ENTRYPOINT` to get rid of it)

PR #42, opened 2015-02-06
Add CGO-enabled golang-cross images

attilaolah:
This adds two new tags, `1.3.3-cross-cgo` and `1.4.2-cross-cgo`. They build on top of `*-cross`, with the addition of various `crossbuild-essential-*` packages that allow cross-compiling CGO code by setting e.g. `CC=arm-linux-gnueabihf-gcc`.

This is just a proposal but I figured it might be useful for someone else as well.

### DIFF
--- a/1.3/cross-cgo/Dockerfile
+++ b/1.3/cross-cgo/Dockerfile
@@ -1,0 +1,29 @@
+FROM golang:1.3.3-cross
+
+# For a list of supported platforms, see:
+# https://wiki.debian.org/CrossToolchains#For_jessie.2Ftesting
+# Some platforms are skipped due to missing dependencies:
+# - mips: missing crossbuild-essential-mips
+# - powerpc: buggy libc6_2.19-13_powerpc.deb
+# - ppc64el: missing g{cc,++}-powerpc64le-linux-gnu
+ENV DEBIAN_CROSSPLATFORMS \
+	arm64 armel armhf mipsel
+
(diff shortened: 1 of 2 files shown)


luxas:
@attilaolah I would like to thank you for this PR.
This helped me a lot. I'm now compiling dynamic ARMv6 binaries on `amd64` hosts without problems

>  it might be useful for someone else as well.
Yes, very

Have a good day!

tianon:
From http://emdebian.org/:
> As of July 2014, updates to the Emdebian distributions ceased. There will be no further updates and no further stable releases.

:cry:

luxas:
Thanks @tianon for the valuable information.
I'm about to include something like this in kubernetes/kubernetes#19769
Do you have any ideas how to include this anyway? Ubuntu crosstools here: https://wiki.ubuntu.com/ToolChain#Cross_development_toolchain

But the `emdebian` packages do work (even though they are deprecated), and they're updated `	14-Mar-2015` for jessie

@tianon If e.g. the `ubuntu` (or `emdebian`) packages work nicely, could they be integrated into something like `golang:1.6.1` or `golang:1.6.1-cgo`?

tianon:
Sorry, going back through old issues -- given that `cross` isn't a separate image anymore, and given the complexity involved in implementing this, I still do not think this makes sense for the official images.

However, I do think it's a really cool effort, and should continue as a project outside the official images if someone is interested in taking it up! (Maybe by now several folks already have?)

PR #44, opened 2015-02-26
Make $GOPATH permissive to enable running as non-root

### LINKED ISSUE #43: Should $GOPATH be owned by root?
Docker's [best practices states](https://docs.docker.com/articles/dockerfile_best-practices/#user):

> If a service can run without privileges, use USER to change to a non-root user.

I'm wondering if we should make a non-root user to own `/go` and to not open this security can of worms by default.

yosifkit:
fixes #43


tianon:
LGTM

PR #61, opened 2015-08-31
Add 1.5-wheezy version

Luit:
I'm not sure if what's in this PR is enough to add this version, but the intention was to add a 1.5-wheezy build. 

I would love to have this because we build a Go application that will run on Debian Wheezy, and we like to build a .deb package using Docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ sudo: 9000
 
 env:
   - VERSION=1.5 VARIANT=
+  - VERSION=1.5 VARIANT=wheezy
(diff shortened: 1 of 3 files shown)


yosifkit:
Seems fine to me. Could you add this new variant to the [`.travis.yml`](https://github.com/docker-library/golang/blob/master/.travis.yml#L7)?

Luit:
Like so?

yosifkit:
:+1: LGTM

tianon:
LGTM

PR #66, opened 2015-09-17
Add "g++" (since Go technically does support it)

tianon:
Also, reorganizes the way we install the deps to match other official images slightly better.


tianon:
Whoo, we're green! :+1:

tianon:
(For proof, see https://golang.org/cmd/cgo/, Ctrl+F for "C++".)

PR #74, opened 2015-12-30
Update to Alpine 3.3 (especially for "--no-cache") and make use of that sweet "--virtual" feature of apk

tianon:


yosifkit:
LGTM

PR #76, opened 2016-01-07
add nsswitch.conf to support /etc/hosts for --net and --link

Jackong:


tianon:
Sorry for the delay, but can you elaborate a bit on what problem this is solving and why it's necessary/proper?  It feels a little strange to be providing `nsswitch.conf` contents that differ from the default provided by our image base (and on top of that, Docker itself has moved away from using `/etc/hosts` and now hijacks DNS more directly, and Alpine 3.4 had several DNS-related fixes as well).

ei-grad:
It looks the way docker hijacks DNS doesn't work, and Golang does its awful deed on behance of GNU trolls, who claim to use DNS before FILES in case if nsswitch is missed. I spent a day debugging this. Burn in hell.

ei-grad:
Is there any way to run test like:

```
→ docker run -it --rm golang:alpine sh
Unable to find image 'golang:alpine' locally
alpine: Pulling from library/golang
3690ec4760f9: Already exists 
c311f66f3ac9: Already exists 
89656d2b09ac: Pull complete 
1602314b9bf7: Pull complete 
181529c471d2: Pull complete 
b561214ca853: Pull complete 
Digest: sha256:c502cbf4608374447cef87d493047ca1a33fc02c0a100f29fe186653881c4b25
Status: Downloaded newer image for golang:alpine
/go # cat > t.go << EOF
package main

import "net"

func main() {
    net.LookupHost("linked")
}
EOF
/go # GODEBUG=netdns=go+2 go run t.go 2>&1 | grep dns,files && echo "Burn in hell!"
go package net: hostLookupOrder(linked) = dns,files
Burn in hell!
```

fredleger:
+1 for this. very anoying for us

tianon:
I'm convinced -- if you'd like to rebase this to the current versions and update it to match what was implemented in https://github.com/docker-library/docker/pull/84, that'd be great, otherwise I'll do so when I get a chance. :+1: :heart:

tianon:
To be more specific:
```dockerfile
# set up nsswitch.conf for Go's "netgo" implementation
# - https://github.com/golang/go/blob/go1.9.1/src/net/conf.go#L194-L275
# - docker run --rm debian:stretch grep '^hosts:' /etc/nsswitch.conf
RUN [ ! -e /etc/nsswitch.conf ] && echo 'hosts: files dns' > /etc/nsswitch.conf
```

PR #85, opened 2016-02-25
Add git to Alpine images

lorenz:
This enables the use of `go get` inside the Alpine images without first installing git.


tianon:
This is a duplicate of #69 -- I think the discussion there is probably going to be interesting to you. :smile:

lorenz:
Makes sense!

PR #91, opened 2016-03-25
add -fno-PIC when using external linker on alpine

ncopa:
The Alpine toolchain enables PIE by default. Since the go linker does
not (yet) support PIE we need to explicitly disable it on Alpine.

This is a workaround for https://github.com/golang/go/issues/14851

### DIFF
--- a/1.6/alpine/Dockerfile
+++ b/1.6/alpine/Dockerfile
@@ -8,6 +8,9 @@ ENV GOLANG_BOOTSTRAP_VERSION 1.4.3
 ENV GOLANG_BOOTSTRAP_URL https://golang.org/dl/go$GOLANG_BOOTSTRAP_VERSION.src.tar.gz
 ENV GOLANG_BOOTSTRAP_SHA1 486db10dc571a55c8d795365070f66d343458c48
 
+# https://golang.org/issue/14851
+COPY no-pic.patch /
+
 RUN set -ex \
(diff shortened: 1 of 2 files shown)


1.6/alpine/Dockerfile
```diff
@@ -30,11 +31,13 @@ RUN set -ex \
...
 	&& rm golang.tar.gz \
 	&& cd /usr/local/go/src \
+	&& patch -p2 -i /no-pic.patch \
 	&& ./make.bash \
 	\
-	&& rm -rf /usr/local/bootstrap /usr/local/go/pkg/bootstrap \
+	&& rm -rf /usr/local/bootstrap /usr/local/go/pkg/bootstrap /*.patch \
 	&& apk del .build-deps
 
+
```
tianon: Can you move this extra blank line up between `ENV` and `COPY` on lines 9 and 10? :pray:  (and please add a comment referencing https://golang.org/issue/14851 while you're at it :+1:)

ie:
```dockerfile
ENV GOLANG_....

# https://golang.org/issue/14851
COPY no-pic.patch /

RUN ...
```

ncopa:
good suggestions :+1: pushed updated fix. thanks!

tianon:
I want a reminder of where we need to look for whether we can take this back out yet. :innocent: :+1:

Thanks for updating!  LGTM

yosifkit:
LGTM

PR #95, opened 2016-06-06
Convert "generate-stackbrew-library.sh" to output the new 2822-based format

tianon:
cc @md5 (relevant to your interests)

It's not Perl, but I think it's easy enough to read anyways (and even in Perl, we'll have the same copy-pasta between repos that we'll have with this).

Once we've got this straw man somewhere we're happy with, I'll be slowly considering converting more repos over with similar PRs.

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -1,42 +1,88 @@
 #!/bin/bash
-set -e
+set -eu
 
 declare -A aliases
 aliases=(
 	[1.6]='1 latest'
 )
 
+self="$(basename "$BASH_SOURCE")"
 cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
 
 versions=( */ )
 versions=( "${versions[@]%/}" )
-url='git://github.com/docker-library/golang'
 
-echo '# maintainer: InfoSiftr <github@infosiftr.com> (@infosiftr)'
-echo '# maintainer: Johan Euphrosine <proppy@google.com> (@proppy)'
+# get the most recent commit which modified any of "$@"
+fileCommit() {
+	git log -1 --format='format:%H' HEAD -- "$@"
+}
+
+# get the most recent commit which modified "$1/Dockerfile" or any file COPY'd from "$1/Dockerfile"
+dirCommit() {
+	local dir="$1"; shift
+	(
+		cd "$dir"
+		fileCommit \
+			Dockerfile \
+			$(git show HEAD:./Dockerfile | awk '
+				toupper($1) == "COPY" {
+					for (i = 2; i < NF; i++) {
+						print $i
+					}
+				}
+			')
+	)
+}
+
+cat <<-EOH
+# this file is generated via https://github.com/docker-library/golang/blob/$(fileCommit "$self")/$self
+
+Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
+             Joseph Ferguson <yosifkit@gmail.com> (@yosifkit),
+             Johan Euphrosine <proppy@google.com> (@proppy)
+GitRepo: https://github.com/docker-library/golang.git
+
+EOH
+
+# prints "$2$1$3$1...$N"
+join() {
+	local sep="$1"; shift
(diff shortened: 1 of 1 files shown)


generate-stackbrew-library.sh
```diff
@@ -10,33 +10,77 @@ cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
...
+Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
+             Joseph Ferguson <yosifkit@gmail.com> (@yosifkit),
+             Johan Euphrosine <proppy@google.com> (@proppy)
+GitRepo: https://github.com/docker-library/golang.git
+
+EOH
+
+# prints "Tags: $1, $2, $3, ..."
+outputTags() {
+	echo "Tags:$(printf ', %s' "$@" | cut -d, -f2-)"
```
md5: I think if you're going to do this, you might as well just define `join($delim, $elem...)` and use `Tags: $(join ', ' "${versionAliases[@]}")`, etc. below for clarity.

tianon: I thought about that, but a generic `join` is slightly more complicated than what this is doing (it's prefixing _everything_ with `, `, and then trimming off the first comma, leaving the first space).  Is there a dead-simple generic `join` in Bash we could rip off? :innocent:

generate-stackbrew-library.sh
```diff
@@ -10,33 +10,77 @@ cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
...
+				toupper($1) == "COPY" {
+					for (i = 2; i < NF; i++) {
+						print $i
+					}
+				}
+			')
+	)
+}
+
+self="$(basename "$BASH_SOURCE")"
```
md5: It might be better to move this up by the `cd` line at the top of the script.

generate-stackbrew-library.sh
tianon: (the "obvious" solution that abuses `IFS` doesn't work since it only uses the first char :cry:)

generate-stackbrew-library.sh
```diff
@@ -10,33 +10,77 @@ cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
...
+cat <<-EOH
+# this file is generated via https://github.com/docker-library/golang/blob/$(fileCommit "$self")/$self
```
md5: Are comments like this actually allowed in the RFC 2822 format or is the parser for these files using a looser definition?

md5: I just found the `CommentStripper` code: https://github.com/docker-library/official-images/blob/dfc5512/bashbrew/go/vendor/src/github.com/docker-library/go-dockerlibrary/pkg/stripper/comments.go

generate-stackbrew-library.sh
md5: Yeah, I did a similar hack in my attempt: `$(IFS=, ; echo "${tags[*]/#/ }")`

i.e. Prepend each tag with a space, then set `IFS` to a comma to join the output with an extra space at the beginning

md5: As for a generic one, this works as long as you don't have `%` in your delimiter: http://stackoverflow.com/a/2317171

md5: This one is interesting too: http://stackoverflow.com/a/23673883

generate-stackbrew-library.sh
tianon: Yeah, I've discussed this a lot with @paultag -- the format we're using is technically only loosely based on RFC 2822 -- it's been called "deb822" or "deb2822" before as well, since it's so common in Debian, but we're still trying to think up a good name that @paultag is happy with. :smile:

generate-stackbrew-library.sh
tianon: Nice, updated with a simplified version of that first one. :+1:

generate-stackbrew-library.sh
tianon: :+1: updated

yosifkit:
LGTM

PR #89, opened 2016-03-08
go-wrapper for alpine variant

xanadros:
Added a posix compliant version of the go-wrapper script.

There were only minor changes to the original script to accommodate array usage under /bin/sh

PR #98, opened 2016-06-28
go-wrapper all the things

yosifkit:
This is building upon the changes in #89 so that the same wrapper script works in all versions.

Closes #89


tianon:
Nice, simple fix! :+1:

LGTM

Issue #97, opened 2016-06-27
Go's TLS cannot find root certs in 1.6.2-alpine


jmhodges:
When trying to port to golang:1.6.2-alpine from golang:1.6.2, the Go HTTP client gets errors like:

    x509: failed to load system roots and no roots provided

when making HTTPS requests.

Minimum case:

In the Dockerfile:

    FROM golang:1.6.2-alpine
    CMD ["sh"]`

And then run `go run foo.go` with a `foo.go` of

    package main
    
    import (
	    "fmt"
	    "net/http"
    )
    
    func main() {
	    _, err := http.Get("https://www.google.com")
	    fmt.Printf("welp %s\n", err)
    }

jmhodges:
I see that there's an attempt to install ca-certificates in the the golang:1.6.2-alpine Dockerfile, but it doesn't seem to be enough. I don't know why. Perhaps, the place where alpine puts that package's contents isn't in the list that Go checks?

yosifkit:
`ca-certificates` gets removed with the rest of the [build-deps](https://github.com/docker-library/golang/blob/13b4447729b4367396148cc18745c9901e5fe4a8/1.6/alpine/Dockerfile#L30).  Looks like we need to specifically install and keep it if that is what is generally desired and expected on the super slim alpine images.

yosifkit:
It seems that I forgot that it is on purpose: https://github.com/docker-library/golang/issues/96#issuecomment-228841794

jmhodges:
Well, for evidence that including them might be nice, I don't have a Go container in production that doesn't do HTTPS requests (for SAAS logs, etc.).

It'd be nice for the certs to be included.

capkurmagati:
@yosifkit The problem to me is that the images changes its behaviour without any notification(e.g. create a tag).
I used to deploy my https services based on alpine images, but it doesn't work now.
From the immutable point of view, I think we need to do something.

yosifkit:
This is a duplicate of https://github.com/docker-library/golang/issues/96.  As I mentioned over there, the `ca-certs` problem stems from the update to `alpine:3.4`.

As for immutable image tags, you should probably look in to pulling images by digest.  A full history of shas for each official image are in the docs: https://github.com/docker-library/docs/blob/master/golang/tag-details.md.

yosifkit:
Closing as duplicate

jmhodges:
[...]

PR #99, opened 2016-06-29
add root certificates to Alpine images

### LINKED ISSUE #96: golang:1.x-alpine have no /etc/ssl
How to reproduce it:
```
$ docker run --rm golang:1.6.2-alpine ls /etc/ssl
ls: /etc/ssl: No such file or directory
```
It also happens with `1.5-alpine` and `1.7-alpine`.
I think it relates to https://github.com/docker-library/golang/commit/f3768925fd0ebeb966c30bc64206b0e6020de66e.

Because I built the image from https://github.com/docker-library/golang/commit/0ce80411b9f41e9c3a21fc0a1bffba6ae761825a and `/etc/ssl` existed.

I am not sure if this is a expected behavior but one builds project which send `https` request will get `x509: failed to load system roots and no roots provided`.

How about adding `RUN apk add ca-certificates` to Dockerfile or using another tag?
I would like to create a pull request.

jmhodges:
Fixes #96 (and the duplicate #97).

This allows Go's HTTP library to make HTTPS client connections with secure certificate validation. Without this change, `net/http` in these images return the error `"x509: failed to load system roots and no roots provided"`.

This adds 0.6MB to the image (as measured with docker history).

### DIFF
--- a/1.5/alpine/Dockerfile
+++ b/1.5/alpine/Dockerfile
@@ -1,11 +1,12 @@
 FROM alpine:3.4
 
+RUN apk add --no-cache ca-certificates
+
 ENV GOLANG_VERSION 1.5.4
 ENV GOLANG_SRC_URL https://golang.org/dl/go$GOLANG_VERSION.src.tar.gz
 ENV GOLANG_SRC_SHA256 002acabce7ddc140d0d55891f9d4fcfbdd806b9332fb8b110c91bc91afb0bc93
 
 RUN set -ex \
-	&& apk add --no-cache ca-certificates \
 	&& apk add --no-cache --virtual .build-deps \
 		bash \
 		gcc \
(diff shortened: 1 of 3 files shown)


1.5/alpine/Dockerfile
```diff
@@ -5,9 +5,9 @@ ENV GOLANG_SRC_URL https://golang.org/dl/go$GOLANG_VERSION.src.tar.gz
 ENV GOLANG_SRC_SHA256 002acabce7ddc140d0d55891f9d4fcfbdd806b9332fb8b110c91bc91afb0bc93
 
 RUN set -ex \
+	&& apk add --no-cache ca-certificates \
```
yosifkit: We should just make this a separate `RUN` above the `ENV` lines.  We are already downloading the apk cache twice and this will let us share the layer between the go versions and might even work for other images from `alpine:3.4`

jmhodges: Will do!

jmhodges:
Feedback handled!

Let me know what you think.

yosifkit:
LGTM

ping @tianon

tianon:
Indeed, `net/http` is a fair argument. :+1:

LGTM, thanks!

jmhodges:
Thank you!

PR #100, opened 2016-07-08
Update to 1.7rc1

pwaller:
This is a careful lexical replacement. I haven't tested, but it should be sound. Based on 13b4447729b4367396148cc18745c9901e5fe4a8.

Ping @tianon.


VojtechVitek:
LGTM

tianon:
:+1:

PR #104, opened 2016-08-03
Install git in alpine images

yunspace:
Currently the golang alpine images do not contain `git`, which comes default in the debian based golang images. 

This makes it impossible to run `go get` with an alpine based image. The availability of `git` and go dependency fetching should be consistent across both alpine and debian based images.


tianon:
This is a duplicate of #69 (and more recently #85); I think the discussion over there is likely going to be of interest to you! :+1:

yunspace:
lol thanks... sorry should have searched first.

PR #106, opened 2016-08-03
Add "windowsservercore" variant

tianon:
Closes #92 (commits included here)

Here's a snippet of the updated `generate-stackbrew-library.sh` output:

```
...

Tags: 1.6.3-alpine, 1.6-alpine, 1-alpine, alpine
GitCommit: 9f666dc2f4f51df564613f787d28b3a2353243e0
Directory: 1.6/alpine

Tags: 1.6.3-windowsservercore, 1.6-windowsservercore, 1-windowsservercore, windowsservercore
GitCommit: a30dac9f2492855f5040cb67de95d7181eebc9b8
Directory: 1.6/windows/windowsservercore
Constraints: windowsservercore

Tags: 1.7rc4, 1.7
GitCommit: cdf73bb71f9a0c0f9aa22beb3d9d34274aaf1b03
Directory: 1.7

...
```


tianon:
I've used the following `Dockerfile` to test this (installing `choco` to install `git` to `go get github.com/constabulary/gb`) successfully:
```dockerfile
FROM golang:windowsservercore

RUN Invoke-WebRequest https://chocolatey.org/install.ps1 -UseBasicParsing | Invoke-Expression

RUN choco install -y -params /GitAndUnixToolsOnPath git.install

RUN go get -v github.com/constabulary/gb/...; \
	gb help;
```

A few size notes:
- `microsoft/windowsservercore`: ~7.764 GB
- `golang:windowsservercore`: ~8.211 GB (so Go itself is ~0.447 GB)
- just the "install `choco`" layer above: ~8.279 GB (so `choco` is ~0.068 GB)
- `git.install`: ~8.808 GB (so Git itself is ~0.529 GB -- yes, bigger than Go)
- `go get .../gb/...`: ~8.864 GB

So I'm very strongly -1 on including Git by default, even though it hampers `go get` usage slightly (especially given that it really isn't all that hard to write those two lines to install `choco` and then `git`).

yosifkit:
I am going to argue for including `git`:

- it is included in all* golang images
  - we even include `bzr`, `mercurial`, and `subversion`
- we include build tools like gcc/g++, and make too

So, are we going to treat the 7.7GB base windows image like the `alpine` tags and force users to sideload the tools they actually need?  Maybe we could skip used tools like `git` in the `nanoserver` images?

Possibly useful links to install git without resorting to choco:
 - http://superuser.com/questions/944576/git-for-windows-silent-install-silent-arguments
 - https://chocolatey.org/packages/git.install#files  (ie, just steal the portions of their script for what we need)

<sub>* except `alpine` based images</sub>

tianon:
Fair point -- so `windowsservercore` is the "full-featured" version, and `nanoserver` is the `slim` variant.  Seems sane to me, and matches how `nanoserver` is generally described. :+1:

(Probably worth noting that we're still working on a way to build official `nanoserver`-based images, which is why they are not included here.)

tianon:
Ok, updated to install Git (without installing `choco`, which makes our `Dockerfile` a fair bit more verbose, but IMO is worth it to leave the unrelated `choco` as a choice for our users instead).

yosifkit:
LGTM

tianon:
Step 1, :white_check_mark: 
Step 2, https://github.com/docker-library/official-images/pull/2051

PR #111, opened 2016-08-28
GoLint utility for Debian-based images

### LINKED ISSUE #102: Add golint to golang image
We use golang through docker extensively, e.g.,
```
docker run -i -t --rm -v $PWD:/go/src/... -w /go/src/... golang go build
```
Typically before compiling our source code with golang, we will invoke golint. Currently the golint is not available in golang images so we have to download golint:
```
go get -u github.com/golang/lint/golint 
```

it would be nice if golang docker image comes with golint installed already.

ybubnov:
Install the golint utility to the Debian-based images to
be able validate the Go code before the building binaries.

The golint utility is not going to be installed into the
Alpine-based images, as their size should be as small as
possible.

closes #102

### DIFF
--- a/1.6/Dockerfile
+++ b/1.6/Dockerfile
@@ -23,4 +23,7 @@ ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
 RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
 WORKDIR $GOPATH
 
+RUN go get -u github.com/golang/lint/golint \
+    && rm -rf "$GOPATH/src"/* "$GOPATH/pkg"
+
 COPY go-wrapper /usr/local/bin/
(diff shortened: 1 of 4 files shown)


bondar-pavel:
Looks good to me.
@tianon Could you please take a look?

tianon:
I saw this, but figured we should finish the discussion over on the issue, so I'll mirror my response from there over here: (https://github.com/docker-library/golang/issues/102#issuecomment-243179490)

> I'm personally very strongly -1 on including this by default -- it's trivial for users to add after the fact with a simple two-line `Dockerfile` (`FROM golang:version`, `RUN go get github.com/golang/lint/golint`), and it's not necessary for most use-cases of this image.
> 
> Combine that with automated builds (https://docs.docker.com/docker-hub/builds/) and repository links (https://docs.docker.com/docker-hub/builds/#repository-links) and it's reasonably easy to have an up-to-date image built `FROM` this one with `golint` pre-installed.
> 
> To elaborate a little bit more, if we add `golint`, what's going to be next?  `godeps`?  Where do we draw the line?

ybubnov:
Closing the pull requests as per discussion in #102.

PR #117, opened 2016-11-02
Change default buildmode to 'pie' for alpine

m5p3nc3r:
Now that golang supports building for 'pie' environments, this patch updates the golang build to make use of this.
The dockerfile has been modified to add GO_LDFLAGS='-buildmode=pie', to ensure that the base go image is compatible with alpine.
There is a hack to change the default buildmode from 'default' to 'pie', but this ensures that when people build applications using this image without specifying a specific buildmode, 'pie' will be used.
I have tested this on aarch64, and it allows me to build functional docker images based on golang without explicitly adding -buildmode=pie to their Dockerfile's.


m5p3nc3r:
Ouch!  Not sure what happened there - the build is complaining that gcc cannot be found, but I can see the package was added earlier through 'apk add ...'
Any ideas what has caused this?

m5p3nc3r:
I have just tried compiling the golang-hello-world test by hand using the golang version built by this Dockerfile.  It does fail with the same error.  I am not understanding why gcc is required for the build though?  I can see that gcc is added to the container to build golang, but then removed before packaging.  So I am not understanding what I have changed that pulls in a dependency on gcc?

Sorry, I am new to docker/golang - I thought this was going to be easy :)

m5p3nc3r:
It looks like building with 'pie' pulls in a dependency on gcc and musl-lib.  I have modified the Dockerfile to keep these in the generated alpine golang image.  I have now tested this, and it works on x86 and aarch64 platforms.

m5p3nc3r:
Ok, am going to abandon this patch.  I think I have worked out the correct solution for the problem.

We need to have both -fno-pic and -no-pie for this to run properly in Alpine.  The problem is that the version of gcc that comes with Alpine 3.4 does not support -no-pie.  When I switch to edge (which supports aarch64), I am able to use -fno-pic and -no-pie to build working executables on x86 and arm platforms.

When Alpine move edge into an official release with the new version of gcc, we need to update the no-pic patch to include no-pie as well.

PR #123, opened 2016-12-01
Copy Alpine community go patch of default buildmode PIE

weixiaowilliam:
Current no-pic.patch disable PIE which is expected by Alpine for security hardening
Copy Alpine community go patch to enable PIE by default
(http://git.alpinelinux.org/cgit/aports/tree/community/go/default-buildmode-pie.patch)


tianon:
See also https://github.com/docker-library/golang/pull/117, which is the last PR which attempted this change.

weixiaowilliam:
I don't think PR:https://github.com/docker-library/golang/pull/117 has really enabled PIE since it use -fno-pic and -no-pie which will tell external linker to disable PIE. In addition,  https://github.com/docker-library/golang/pull/117 has dependence on GCC version. But Alpine community patch really enables PIE which is expected by Alpine for security hardening and has no dependence on GCC version.

tianon:
Sure, but it suffered from the same problem this PR introduces -- `gcc` becomes required for building Go programs in the default case, which is a significant change from upstream. :disappointed:

tianon:
It also has a significant effect on image size (given that `gcc` must be included by default) -- 79+ MB in my testing.

weixiaowilliam:
Yes, enabling PIE will incur dependence on GCC and image size increasing and it's the cost for security hardening. Since GoLang internal linker doesn't support "R_TLS_IE", it will invoke external GCC linker for "pie" buildmode. In additon, musl-dev package is needed by GCC to link with musl libraries.
If you run "apk add --no-cache go" inside Alpine it will also automatically install GCC firstly.

tianon:
This repository aims to provide (first and foremost) a Go experience that matches as closely as possible that which is supported and recommended by upstream (and in the case of Alpine variants, doing so in the most minimal way possible).  Modifying Go to always use GCC for linking does the opposite of that (even if it is something that is preferred by the Alpine community/packagers).

IMO, folks who are looking for this sort of modification are better off using Alpine's Go packages directly, since they are designed/modified specifically to handle this use case.

weixiaowilliam:
1. Current "no-pic.patch" is actually for external GCC linker. If GCC is not expected to be installed then the patch doesn't make sense and should be removed.
2. Besides PIE dependence on GCC, some platforms (such as ARM64 and MIPS64) also requires GCC installed since Go internal linker doesn't support these platforms. Do you have plan of supporting ARM64?

PR #128, opened 2016-12-16
Backport of: https://go-review.googlesource.com/#/c/33106/

weixiaowilliam:


tianon:
(This patch is already included in the 1.8 releases, so is not necessary there.)

PR #139, opened 2017-01-23
Add "xenial" variant.

Tyndie:
- Add Golang 1.6, 1.7, 1.8 Ubuntu Xenial 16 LTS docker build files.
- Include xenial variant in update and generation bash scripts.
- Include xenial variant in travis ci configuration file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,16 @@ services: docker
 
 env:
   - VERSION=1.8 VARIANT=
+  - VERSION=1.8 VARIANT=xenial
   - VERSION=1.8 VARIANT=stretch
   - VERSION=1.8 VARIANT=alpine
   - VERSION=1.7 VARIANT=
+  - VERSION=1.7 VARIANT=xenial
   - VERSION=1.7 VARIANT=wheezy
   - VERSION=1.7 VARIANT=alpine
   - VERSION=1.7 VARIANT=alpine3.5
   - VERSION=1.6 VARIANT=
+  - VERSION=1.6 VARIANT=xenial
   - VERSION=1.6 VARIANT=wheezy
   - VERSION=1.6 VARIANT=alpine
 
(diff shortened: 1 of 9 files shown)


tianon:
I think the discussion over on https://github.com/docker-library/memcached/pull/9 is really relevant here too.

yosifkit:
Thank you for your contribution and time.

Unfortunately this is not something we want to support and maintain. For most users, just having "golang" at a specific version fulfills all their requirements. They don't need to care about the underlying distribution it is running within. If they are running their own application, in python for example, and they are unable to run it in our Debian, Alpine, or Windows Server Core based python images, then they are free to install python and their required dependencies within CentOS, Ubuntu, or other base distribution image.

For many images here are the variants that we maintain with short descriptions and why we have them:

 - `debian`
   - what: the default image that is based on [Debian](https://www.debian.org/) but doesn't usually say `debian` in the name
   - why: the smallest mainline distro using glibc; stable set of libraries provided though `apt`; we have familiarity with Debian; it is similar enough to Ubuntu for most users
 - `slim`:
   - what: a smaller version of the Debian based image; to make it really easy to use, many of the language images include commonly used libraries and development headers; this variant doesn't install them
 - `stretch` (or `wheezy`)
   - what: Debian based image built on the next (or previous) stable [release of Debian](https://www.debian.org/releases/)
   - why: to help transition major base image bumps without affecting users drastically
 - `alpine`
   - what: based on the [Alpine Linux](https://alpinelinux.org/) image (which itself is only about 4-5MB)
   - why: gives users an even smaller image to base their software on; caveat is that not everything is libc agnostic and may not work with musl libc
 - `windowsservercore`
   - what: Windows based image that only runs on Windows; based on [`microsoft/windowsservercore` image](https://hub.docker.com/r/microsoft/windowsservercore/)
   - why: allow users to run software written for Windows in a container
 - `nanoserver`
   - what: smaller Windows based image that only runs on Windows; based on [`microsoft/nanoserver`](https://hub.docker.com/r/microsoft/nanoserver/) (base image ~333MB vs ~5GB)

PR #144, opened 2017-02-09
Remove "wheezy" variant from Go 1.8+

### LINKED ISSUE #143: drop wheezy images for newer go versions
I propose we drop the wheezy images;  maybe even as soon as the pending `1.8` release.

It seems that the reason we kept `wheezy` was to have an image based on Debian stable: https://github.com/docker-library/golang/pull/21#issuecomment-58589491.

tianon:
Closes #143


yosifkit:
Build failure is disk space issue on Travis, so LGTM.

PR #147, opened 2017-02-12
upgrade to 2.11.1 and use mingit

friism:


1.6/windows/windowsservercore/Dockerfile
```diff
@@ -4,51 +4,32 @@ FROM microsoft/windowsservercore
...
 	\
 	Write-Host 'Updating PATH ...'; \
-	$env:PATH = 'C:\git\bin;C:\git\mingw64\bin;C:\git\usr\bin;' + $env:PATH; \
+	$env:PATH = 'C:\git\cmd;C:\git\mingw64\bin;C:\git\usr\bin;' + $env:PATH; \
 	[Environment]::SetEnvironmentVariable('PATH', $env:PATH, [EnvironmentVariableTarget]::Machine); \
 	\
 	Write-Host 'Verifying install ...'; \
 	Write-Host '  git --version'; git --version; \
-	Write-Host '  bash --version'; bash --version; \
-	Write-Host '  curl --version'; curl.exe --version; \
```
yosifkit: It seems that the only difference is the loss of bash and curl.  I'm going to guess that those are hardly used anyway in the Windows based image?  Are there any other important changes between the versions git and mingit that could affect users?

friism: That's my analysis also. More details here: https://blogs.msdn.microsoft.com/visualstudioalm/2016/09/03/whats-new-in-git-for-windows-2-10/

The other stuff that's not there is GUI tools (like `git gui` and `gitk`).

I think mingit is a good match for cli-based workflows like this.

yosifkit:
>It currently requires only ~45MB on disk.

Oooooh :heart: :heart_eyes: 
LGTM

1.6/windows/windowsservercore/Dockerfile
yosifkit: Maybe we should link to that blog in the Dockerfile so it is slightly easier to see why we switched to MinGit over regular Git for Windows.

tianon: Agreed -- I'll add a comment in just a second. :+1: 

Thanks @friism :heart: :heart:

friism:
Awesome!

PR #149, opened 2017-02-16
Update to 1.8

belak:
1.8 was officially released, so I moved 1.8-rc to 1.8, ran `update.sh 1.6 1.7 1.8` and updated generate-stackbrew-library.sh to tag 1.8 as 1 and latest


tianon:
LGTM :heart:

Issue #119, opened 2016-11-17
'make' on alpine


atomi:
any thoughts on adding make on the alpine tag considering it wont add much to the size and the wheezy tag includes it as well.

tianon:
For many/most users, it isn't necessary, and it cannot be removed after the fact, so we intentionally try to keep the Alpine image as slim as possible.

See also https://github.com/docker-library/docs/blob/7a1edc542ec7d17626ec5cee87e87ed99da71afb/golang/README.md#golangalpine, specifically:

> To minimize image size, it's uncommon for additional related tools (such as `git` or `bash`) to be included in Alpine-based images. Using this image as a base, add the things you need in your own Dockerfile (see the [`alpine` image description](https://hub.docker.com/_/alpine/) for examples of how to install packages if you are unfamiliar).

atomi:
Yeah I realize keeping the alpine image as slim as possible would actually make it _more_ useful to the __most__ people. It was a bit selfish of a request. Thanks for the response!

tianon:
Your understanding is appreciated :heart: :heart:

taadis:
For
> To minimize image size, it's uncommon for additional related tools (such as git or bash) to be included in Alpine-based images. Using this image as a base, add the things you need in your own Dockerfile (see the alpine image description for examples of how to install packages if you are unfamiliar).

I think the *minimize image size* must be whole, but git deps not exitis in `golang:alpine`, it is not whole, because go get not work in `golang:alpine`.
so, suggest add git deps in `golang:alpine`.

miry:
The sample golang Dockerfile: https://github.com/pubnative/docker-images/blob/master/oncall-sync/Dockerfile

You see, we use mostly golang only for build binaries and uses pure alpine to run it. So it should be as @taadis be whole as golang tool and have similar packages as other builds in golang dockers tags.

PR #157, opened 2017-04-05
Added git, so 'go get' works with git deps

peterleeau:
Currently the alpine image doesn't include git, so following will happen if the dependencies a in a git repo:

go: missing Git command. See https://golang.org/s/gogetcmd


peterleeau:
I seem to step on an existing argument:https://github.com/docker-library/golang/issues/119#issuecomment-266238703

peterleeau:
Close the request to keep it slim.

PR #158, opened 2017-04-25
Convert to templates and add some initial multi-architecture support

tianon:
For building on amd64, this should be 100% equivalent.  What this opens up is the possibility of building these as-is on other architectures (I've tested successfully on s390x by only modifying the `FROM` of `1.8/Dockerfile` and `1.8/alpine/Dockerfile`).


tianon:
"No `FROM` lines were harmed in the making of this PR."

PR #159, opened 2017-04-26
Fix cross-building (by setting GOOS, GOARCH, GO386, and GOARM to our bootstrap Go's values)

tianon:


tianon:
Oh shoot, I think I need to set the `GOHOST*` variants of these too ("`##### Building go_bootstrap for host, linux/amd64.`").

tianon:
For reference, here's the error message we get without this (given that it tries to build host tools for `linux/amd64` with a 32bit/`386` compiler):
```
cc1: sorry, unimplemented: 64-bit mode not compiled in
```

tianon:
Success (`FROM i386/alpine:3.5`):
```
+ export PATH=/usr/local/go/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+ go version
go version go1.8.1 linux/386
 ---> b7b8398aaf95
Removing intermediate container 9e25de2d66d1
```

PR #165, opened 2017-06-14
Add go1.9beta1

tianon:
https://golang.org/dl/#go1.9beta1
https://tip.golang.org/doc/go1.9


yosifkit:
go 1.9 build successful!

PR #166, opened 2017-06-16
add make in alpine

tiny1990:
add make in alpine is better for us


tianon:
This is a duplicate of https://github.com/docker-library/golang/issues/119 -- please read the discussion there for why this is intentionally not included. Thanks! :+1:

PR #173, opened 2017-08-25
Fix typo in go-wrapper description

drornir:


tianon:
Nice catch!  Want to update all the other copies of this file in the repo as well?

drornir:
sure

PR #175, opened 2017-09-05
Add onbuild image for golang 1.9

djeikyb:


tianon:
This was left out intentionally given that `onbuild` image variants are officially deprecated.  See https://github.com/docker-library/official-images/issues/2076 and https://github.com/docker-library/docs/pull/911.

djeikyb:
@tianon oh okay. there's no mention of that in this project's readme, or https://hub.docker.com/_/golang/ . Because there was no mention of this, I thought the missing `onbuild` image was an oversight rather than intentional. There's no need to dredge up the decision to nix it, but in lieu of this PR, does it make sense to add documentation to https://hub.docker.com/_/golang/ ?

tianon:
Ah, it not being mentioned on https://hub.docker.com/_/golang/ is an accidental side-effect of the `golang:onbuild` tag going away -- normally, when an image has an `onbuild` tag, the contents of https://github.com/docker-library/docs/blob/6b6b3f34023ab90821453ed1e88e7e9165c6b0d1/.template-helpers/variant-onbuild.md get added automatically, which describes the deprecation.  I think we might just need to adjust the way we check variants in our docs scripts to properly pull that in even if we only have `1.8-onbuild`.  I'll take a stab at that and ping you on the resulting PR -- sound good? :pray: :innocent:

djeikyb:
Sounds great! Because the deprecation was above the scope of the golang docker images, I wasn't sure if it was even appropriate to document there. But after golang 1.9 was released I did wait a few days for the onbuild image..lol

PR #181, opened 2017-10-25
Ppc64 fixes

kolyshkin:
This includes two patches. First is required for both 1.8 and 1.9 for Alpine (without it, `go <cmd>` just segfaults). Second is needed for Go 1.9 and 1.9.1 on all distros.


tianon:
Nice, already applied upstream too! :+1: :heart:

Issue #184, opened 2017-11-12
arm32v6 image failing to build on RPi ZeroW - Illegal instruction (core dumped)


rgee0:
Trying to perform a multi stage build from `arm32v6/golang:1.9.2-alpine3.6` on an RPi ZeroW results in a core dump at the first `go` call.

OS:
```
cat /etc/os-release
PRETTY_NAME="Raspbian GNU/Linux 8 (jessie)"
NAME="Raspbian GNU/Linux"
VERSION_ID="8"
VERSION="8 (jessie)"
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"
```

Docker:
```
Client:
 Version:      17.10.0-ce
 API version:  1.33
 Go version:   go1.8.3
 Git commit:   f4ffd25
 Built:        Tue Oct 17 19:11:55 2017
 OS/Arch:      linux/arm

Server:
 Version:      17.10.0-ce
 API version:  1.33 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   f4ffd25
 Built:        Tue Oct 17 19:05:34 2017
 OS/Arch:      linux/arm
 Experimental: false
```

Dockerfile:
```dockerfile
FROM arm32v6/golang:1.9.2-alpine3.6 AS build

ENV GOPATH=/go

RUN go get -d -u github.com/alexellis/blinkt_go/sysfs/gpio

RUN mkdir -p /go/src/github.com/alexellis/blink_go_examples/example
WORKDIR /go/src/github.com/alexellis/blinkt_go_examples/example

COPY . .

RUN GOARM=6 CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o /usr/bin/app .

FROM scratch

COPY --from=build /usr/bin/app /app

CMD ["/app"]

```

Output from `docker build .`
```
Sending build context to Docker daemon  8.192kB
Step 1/9 : FROM arm32v6/golang:1.9.2-alpine3.6 AS build
1.9.2-alpine3.6: Pulling from arm32v6/golang
Digest: sha256:2075c33b0adb86a5625efae7676296ab1d9a08bdd1b0a7308a89080583f18901
Status: Downloaded newer image for arm32v6/golang:1.9.2-alpine3.6
 ---> 14f40308ee24
[...]

PR #187, opened 2017-11-27
Explicitly set GOARM and GO386 to their appropriate values for our target architecture

tianon:
This is necessary because in Alpine's `go` package, using `go env GOARM` and `go env GO386` both always come back with the empty string.

I _think_ this is the proper fix for https://github.com/docker-library/golang/issues/184, but I'm not sure (and don't have a Pi 1 or zero to verify with).  Regardless of whether it fixes that, this is a correct change we should make. :+1:


tianon:
Travis is green, merging. :+1:

PR #192, opened 2017-12-09
Add golang:1.9-alpine-3.7

sapk:
This would provide a stable golang image on alpine3.7 in the wait of go1.10


sapk:
```
$ git diff --no-index 1.9/alpine3.6 1.9/alpine3.7
diff --git a/1.9/alpine3.6/Dockerfile b/1.9/alpine3.7/Dockerfile
index f9a6904..77aa5ef 100644
--- a/1.9/alpine3.6/Dockerfile
+++ b/1.9/alpine3.7/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.6
+FROM alpine:3.7
 
 RUN apk add --no-cache ca-certificates
 
```

generate-stackbrew-library.sh
```diff
@@ -13,7 +13,7 @@ declare -A debianSuite=(
 defaultAlpineVersion='3.7'
 declare -A alpineVersion=(
 	[1.8]='3.5'
-	[1.9]='3.6'
+	[1.9]='3.7'
```
tianon: This one should be left at 3.6 -- this controls what `golang:1.9-alpine` points to, which needs to stay at 3.6.

sapk:
Thanks to you for the ultra quick review. 
If I am not mistaken https://github.com/docker-library/docs/blob/master/golang/README.md is updated automatically ?

tianon:
Yeah, once the image is actually updated (which will likely happen on
Monday).

PR #205, opened 2018-01-25
Deprecate "go-wrapper" script in 1.10+

tianon:
This was a custom script we added way back when in order to support `golang:onbuild`, which as of 1.9 is officially gone.  It hasn't been touched in any major way since Go 1.4 (when we added "custom import path" support).


tianon:
See also https://github.com/docker-library/docs/pull/1128.

jstangroome:
Updating my existing Dockerfiles (as per https://github.com/docker-library/docs/pull/1128 ) to use `FROM golang:1.10` instead of `FROM golang:1.9` and to use `RUN go install -v ./...`  instead of `RUN go-wrapper install` as per this change to the docs, now breaks the `docker build` with:

```
can't load package: package app: code in directory /go/src/app expects import "github.com/jstangroome/what-how"
```

See gist reproducing issue with both previously working and now broken Dockerfiles:
https://gist.github.com/jstangroome/93742d47d8b4965877a924fc99f60c9e

tianon:
Correct -- if you are using a custom import path (or canonical import path, as upstream calls them), then instead of `/go/src/app` in your workdir, the Go tooling _requires_ that you place your source in the correct place within `GOPATH`, which in your case is `GOPATH/src/github.com/jstangroome/what-how`, or `/go/src/github.com/jstangroome/what-how` (which is a detail that `go-wrapper` papered over for you before now -- see all the hacky code in https://github.com/docker-library/golang/blob/eee7a0473995ff99366babe918f0d7d63ab0d741/go-wrapper#L59-L73 which creates a symlink, which the Go upstream team officially also doesn't actually support, so it's really a mircle that worked before).

jstangroome:
@tianon Thanks for clarifying. I was familiar with the "hacky" symlinking done by go-wrapper and quite liked it. I was expecting I'd have to begin doing the symlinking myself now in the `golang:1.10` image. Now that I know it is unsupported, I'll resort to app-specific paths per Dockerfile going forward.

PR #212, opened 2018-03-06
LICENSE 2018

sakajunquality:
update LICENSE to the latest

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2014-2015 Docker, Inc. All rights reserved.
+Copyright (c) 2014-2018 Docker, Inc. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


yosifkit:
Seems reasonable (though also silly seeing how git tracks all changes so every commit means the project was "published" again and so it should always be the current year).  Useful info: https://ben.balter.com/2015/06/03/copyright-notices-for-websites-and-open-source-projects/

cc @tianon

tianon:
Yeah, I tend to agree -- updating copyright ending years seems like busywork.  It's my understanding that generally, the meaningful bit of information is the starting year.

sakajunquality:
@tianon @yosifkit 
Just left the year of beginning

https://opensource.org/licenses/BSD-3-Clause

tianon:
Thanks!

PR #213, opened 2018-03-17
Remove obsolete no-pic.patch from alpine images

nogoegst:
It was necessary for Go < 1.9. Now there are only 1.9 and 1.10 so we don't need this patch anymore.


tianon:
I've successfully verified this against `github.com/mattes/migrate/cli`, which is the example that was failing in https://github.com/golang/go/issues/14851. :+1:

(For future reference, this was added in https://github.com/docker-library/golang/pull/91.)

tianon:
(Once Travis is happy, we're good to merge. :+1:)

PR #215, opened 2018-03-30
Set windows tls version for download from github

ebati:
Currently windows server builds fail because github only support Tls 1.2 but windows server defaults to tls 1.0 while downloading git-for-windows.


parkr:
👍 https://githubengineering.com/crypto-removal-notice/

tianon:
This matches https://github.com/docker-library/python/pull/228; LGTM. :+1: :heart:

PR #217, opened 2018-05-01
Add nsswitch.conf to Alpine images for "netgo"

tianon:
Closes #76

### DIFF
--- a/1.10/alpine3.7/Dockerfile
+++ b/1.10/alpine3.7/Dockerfile
@@ -1,6 +1,12 @@
 FROM alpine:3.7
 
-RUN apk add --no-cache ca-certificates
+RUN apk add --no-cache \
+		ca-certificates
+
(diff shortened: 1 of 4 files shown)


tianon:
This should match https://github.com/docker-library/docker/pull/84. :+1:

1.10/alpine3.7/Dockerfile
```diff
@@ -1,6 +1,12 @@
 FROM alpine:3.7
 
-RUN apk add --no-cache ca-certificates
+RUN apk add --no-cache \
+		ca-certificates
+
+# set up nsswitch.conf for Go's "netgo" implementation
+# - https://github.com/golang/go/blob/go1.9.1/src/net/conf.go#L194-L275
```
tianon: Appears unchanged in 1.10.2 (which is a good sign for this PR). :+1:

PR #229, opened 2018-07-11
Deprecate Alpine 3.6 support

tianon:
We have Alpine 3.7 and 3.8 now, and it's only Go 1.9 that even had 3.6, so let's just remove it and move on (especially given the current `wget` failure it has).


generate-stackbrew-library.sh
```diff
@@ -11,7 +11,7 @@ declare -A debianSuite=(
 )
 defaultAlpineVersion='3.8'
 declare -A alpineVersion=(
-	[1.9]='3.6'
+	#[1.9]='3.7'
```
tianon: "Sneaking" in this bump we probably should've done a little while ago too (making `golang:1.9-alpine` update to the latest Alpine release now).

yosifkit: image

You weren't sneaking!

PR #232, opened 2018-08-17
feat: remove unused files

orisano:
remove unused files to reduce the size of the layer.
389MB -> 255MB

### DIFF
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -51,6 +51,8 @@ RUN set -eux; \
 	done; \
 	./make.bash; \
 	\
+	rm -rf /usr/local/go/pkg/bootstrap; \
+	rm -rf /usr/local/go/pkg/obj/go-build/; \
 	rm -rf /go-alpine-patches; \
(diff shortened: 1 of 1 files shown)


tianon:
Are these directories documented somewhere as being safe to remove post-build? (so we can link back to something to note why we do so)

orisano:
about: /usr/local/go/pkg/bootstrap
https://github.com/golang/go/blob/master/src/cmd/dist/buildtool.go#L121-L125

orisano:
about: /usr/local/go/pkg/obj/go-build/
it is a cache directory.
https://groups.google.com/forum/#!topic/golang-dev/qfa3mHN4ZPA

related issue:
https://github.com/golang/go/issues/23299
https://go-review.googlesource.com/c/build/+/82095

yosifkit:
> cmd/release: delete go/pkg/obj
> 
> There's no point to keeping this in releases.
> It probably shouldn't be created at all, but since it is,
> delete it.
>
> https://go-review.googlesource.com/c/build/+/82095

:smile:

PR #239, opened 2018-09-23
Add git in order to be able to run `go get`

jcabmora:


jcabmora:
Just noticed [this comment](https://github.com/docker-library/golang/issues/119#issuecomment-266238703). I think from a usability perspective, it does make sense to include git. Otherwise, you will keep having people running into the same issue that I (and from the history of PRs) other people have. Yes, you add about 15MB more to the image, but I think it is totally worth it.  I think you have to analyze what the target audience for this image is, I might be wrong but I would not use this image for distributing the apps that I am building, instead it is awesome resource to compile go sources. I was looking into this image so I can provision build agents quickly, and the alternative of using the Ubuntu base image is not really a good one (750 MB!).

Issue #250, opened 2018-11-13
Tag request: alpine with git


JorritSalverda:
Although there's been a number of issues filed on git missing in the alpine images, it would be nice to have a dedicated alpine-git tag that does include git. For runtime the regular alpine image should indeed be kept as small as possible without git installed. But we use the alpine images to build our code and particularly with the new modules system it's very useful to have git installed.

The official docker images have similar tagged image with git preinstalled, see https://hub.docker.com/r/library/docker/tags/.

What do you think? Should be a relatively simple addition from your side, right?

yosifkit:
We are very reluctant in adding more variations to images since it increases the build times and can delay releases and base image induced rebuilds.

For the `docker` images it made sense at the time to make a separate tag for adding git.  Currently, I'd be more in favor of removing the `git` version since it is not necessary for the docker cli binary to run and it is for a feature that is rarely used.  For reference, the `git` tag was added in 2015 (https://github.com/docker-library/docker/issues/3).

I still see this as a duplicate of #119, #140, #153, #157, #166, #198, #209, #230, #239, #241, #251.

david-l-riley:
What's the objection to adding this with `apk`? Not all environments need git (even for Go), and you're likely to eventually need other customizations to your environment (e.g. `gcc` and `musl-dev` if you need `cgo` to work, `protobuf-dev` if you need GRPC, etc.). Alpine is intentionally very slim with a good package manager.

JorritSalverda:
Of course it's possible to install it with `apk`, which is what I already do. However to save time on every build - I use this image only for building, for the runtime container I use `scratch` as most people should do if they can - and not need to keep updating my own builder image for every minor release it would be nice for the official image to also have an _alpine_ version with `git` pre-installed. The regular _alpine_ version should indeed be kept as small as possible.

The reason to require `git` is really due to the golang toolchain requiring it for commands like `go get` and the experimental package versioning system. It's essentially a part of golang's package management system and hence it's like creating a `node` container without `npm` installed.

In the mean time I'll look at better ways to cache containers locally in my [Estafette CI](https://estafette.io) build system so I can use the _Debian_ version instead; now it's just too big and slow to pull in on every build.

gsf:
One use case is unit tests. Helps to have a standard image that will run quick.

tianon:
Ok, let me try to reframe this a little bit.

[...]

PR #251, opened 2018-11-13
Added git and openssh to Dockerfile

thezillion:
Git is required for `go get github.com/package-name` to work.

Added git to all present Dockerfiles.


tianon:
See #239 (for a recent example this is a duplicate of) and https://github.com/docker-library/golang/issues/119#issuecomment-266238703 for a bit more explanation.

thezillion:
Maybe add a git tag like `golang:alpine-git` ?

tianon:
See also https://github.com/docker-library/golang/issues/250#issuecomment-438486997

PR #256, opened 2018-12-19
Add 1.12beta1

tianon:


generate-stackbrew-library.sh
```diff
@@ -3,6 +3,7 @@ set -Eeuo pipefail
 
 declare -A aliases=(
 	[1.11]='1 latest'
+	[1.12-rc]='rc'
```
thaJeztah: Should these be `1.12beta1`, instead of `rc`? (I see some other locations above as well where `-rc` is used)

tianon: We use 'rc' as a generic "pre-release" designation in folder names and tags; this should generate the beta 1 tag too.

thaJeztah: ah, thanks! sorry, I got confused 😊

PR #257, opened 2018-12-22
Add support for Windows Server 1809

kvonbredow:
Adding support for the new 1809 images, and following the base images to mcr.microsoft.com as described in the blog post here: https://azure.microsoft.com/en-us/blog/microsoft-syndicates-container-catalog/

### DIFF
--- a/1.10/windows/nanoserver-sac2016/Dockerfile
+++ b/1.10/windows/nanoserver-sac2016/Dockerfile
@@ -1,4 +1,4 @@
-FROM microsoft/nanoserver:sac2016
+FROM mcr.microsoft.com/windows/nanoserver:sac2016
 
 # $ProgressPreference: https://github.com/PowerShell/PowerShell/issues/2138#issuecomment-251261324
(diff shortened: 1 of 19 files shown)


tianon:
This registry change is really a strange/confusing move by Microsoft IMO. :disappointed:  (They were already hosting the layers, why add additional user-facing redirection?  What's the benefit to users?)

I tried setting up a new Jenkins build worker to support this on Server 2019, but it appears there have been some changes that I haven't had a chance to dig into yet such that with a default install of OpenJDK just like we've done on previous Semi-Annual and LTSC releases causes errors with the Jenkins agent starting.

There's also the issue of LTSC 2019 and 1809 being the same thing in the Docker images (since they're core-only and that's the primary differentiator), so we need to decide whether we're going to call it `1809` or whether we'll use `ltsc2019` instead.  I am leaning towards `1809` since it matches our majority better.

image

tianon:
Ok, given https://github.com/docker-library/official-images/pull/5239 and the 2019 build server I managed to get working, I think this is good to go. :metal:

tianon:
See also https://github.com/docker-library/hello-world/pull/54, which might be worth putting through first/concurrently.

PR #289, opened 2019-07-09
 Add Windows Server Core version 1903 aka May 2019 Update.

olegburov:
Adding support for Windows Server Core version 1903 aka May 2019 Update.

### DIFF
--- a/1.11/windows/windowsservercore-1903/Dockerfile
+++ b/1.11/windows/windowsservercore-1903/Dockerfile
@@ -1,0 +1,73 @@
+FROM mcr.microsoft.com/windows/servercore:1903
+
+# $ProgressPreference: https://github.com/PowerShell/PowerShell/issues/2138#issuecomment-251261324
+SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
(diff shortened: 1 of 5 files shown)


tianon:
We don't currently have a place for these to build, unfortunately (we currently use a Windows Server 2019 VM which is nice because it can actually be updated properly, but has the downside that semi-annual releases like 1903 don't get included). :disappointed:

drpebcak:
This would be nice to have... Any idea what is going to happen with Windows support going forward?

tianon:
These semi-annual releases are honestly a huge pain for us given that they require building on said semi-annual release (which is a pain to run/manage) *or* Windows 10 (which we don't have a license to run in our current platform).

Throwing out the big dreams, I'd love to see the Windows kernel version mismatch not matter anymore, but I realize that's an architectural nightmare for the Windows folks. :disappointed:

drpebcak:
Well, in 3 more months there will be another one. Is there anything that can be done to help with this? In the meantime, I've just forked this repo and applied the patch to build my own, but long-term it would be nice to have this in the upstream.

What platform are you using that you don't have an 1903 image available to spin up real quick for builds? What is the pain point for you guys to spin that up for a build?

PR #294, opened 2019-08-08
Add newer Nano Server variants

tianon:
Closes #280

PR #295, opened 2019-08-13
Update to golang 1.11.13 and 1.12.8

pgporada:
New versions are up per https://golang.org/dl/ to address security vulnerabilities


tianon:
Thanks!

Normally this would've been picked up by https://doi-janky.infosiftr.net/job/update.sh/job/golang/, but as seen in https://doi-janky.infosiftr.net/job/update.sh/job/golang/4302/console we got caught by the combination of the "unknown" problems discussed in https://github.com/docker-library/official-images/issues/3383 plus https://github.com/docker-library/golang/pull/294. :disappointed:

PR #301, opened 2019-09-07
Windows: update MinGit to 2.23.0

thaJeztah:
Git/MinGit was last updated in f6c1fd8462d5b33c7b984cfb99372f122df21d9d (https://github.com/docker-library/golang/pull/147) (over two years ago),
and comparing with the Linux variants, the Windows version was quite a bit behind, so it's probably
a good thing to update this package to get bug fixes in, and bring it more in line with the
Linux variants;

```bash
docker run -it --rm golang:1.13 git --version
git version 2.20.1

docker run -it --rm golang:1.13-alpine sh -c 'apk add --no-progress --no-cache --quiet git && git --version'
git version 2.22.0
```

full diff: https://github.com/git-for-windows/git/compare/v2.11.1.windows.1..v2.23.0.windows.1
Release notes: https://github.com/git-for-windows/git/releases/tag/v2.23.0.windows.1

### DIFF
--- a/1.12/windows/windowsservercore-1803/Dockerfile
+++ b/1.12/windows/windowsservercore-1803/Dockerfile
@@ -7,10 +7,10 @@ SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPref
 # https://blogs.msdn.microsoft.com/visualstudioalm/2016/09/03/whats-new-in-git-for-windows-2-10/
 # "Essentially, it is a Git for Windows that was stripped down as much as possible without sacrificing the functionality in which 3rd-party software may be interested."
 # "It currently requires only ~45MB on disk."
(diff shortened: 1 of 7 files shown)


thaJeztah:
@tianon ptal 🤗 

/cc @ddebroy @vikramhh ptal

thaJeztah:
FWIW; I noticed they now also have a busybox based variant;
Screenshot 2019-09-07 at 15 53 05

Extracting that, it looks like the size of that package is roughly 8MB smaller

Screenshot 2019-09-07 at 15 53 28

Compared to the size of the Windows images themselves I'm not sure if that's worth the risk switching to (it may be risky as busybox may at times be slightly "different")

tianon:
Nice, thanks!

That BusyBox thing definely looks interesting, but for sure something to explore separately. 😅

PR #300, opened 2019-09-05
Add stretch variant for 1.13

tianon:
@leighmcculloch made a pretty compelling argument for this IMO :smile: :innocent:

Closes #296

PR #306, opened 2019-10-16
Add Jessie variant for 1.13 and 1.12

simPod:
_I created those images because I needed them for internal project._

Similar to https://github.com/docker-library/golang/pull/300. Opening PR here if interested in adding Jessie which still has LTS support for now. If not, close :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ matrix:
     - os: linux
       env: VERSION=1.13 VARIANT=stretch
     - os: linux
+      env: VERSION=1.13 VARIANT=jessie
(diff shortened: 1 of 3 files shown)


tianon:
I'm frankly somewhat more -1 on adding *new* Jessie variant images given that Jessie is not officially supported by the Debian Security Team anymore (https://wiki.debian.org/LTS).

simPod:
@tianon kk, let's close this and pretend it never happened 😶

tianon:
No worries, it's a fair ask! :+1: :heart:

Issue #204, opened 2018-01-25
[feature request]: Please add the dep tool to the image


agnivade:
Currently, for CI purposes, one needs to additionally download the `dep` tool. It would be great if it was packaged by default with the image.

FWIW - `yarn` is packaged inside the `node` image.

tianon:
I'm inclined to _not_ include this for the reasons pointed out by the `dep` project:

https://github.com/golang/dep

> `dep` is the official _experiment_, but not yet the official tool. Check out the [Roadmap](https://github.com/golang/dep/wiki/Roadmap) for more on what this means!

https://github.com/golang/dep/wiki/Roadmap

> The goal with `dep` is to be absorbed into the `go` toolchain.

For users who want to play with it before it's officially included in the standard distribution of Go (which will make it naturally available here with no functional changes to the structure of the image), it should be fairly trivial to grab it from https://github.com/golang/dep/releases or via `go get -u github.com/golang/dep/cmd/dep` for those feeling adventurous enough to use master.

What I don't want to do is inadvertently discourage folks from trying `dep` alternatives because we chose a winner before it was actually official, and we're definitely not going to include `godep` and/or other alternatives.

In the case of the `node` image, NPM is included in the standard distribution of Node.js, and it was a decision of the official upstream Node.js working group responsible for the Docker images to include Yarn as well.

In our case, we don't speak for the Go community at all -- we're simply packagers.

agnivade:
FWIW, `dep` is more or less the standard package manager in the Go community. Even other package managers suggest to use `dep`. _Until_ it is merged with the toolchain, this is the recommended tool to use. 

Yes, it is trivial to to grab and install it. That is what I am doing currently. But small things like these make our lives just a little bit easier. :smile: 

But at the end of the day, you are the stakeholder of the project. And I think you have already made your decision. In that case, feel fee to close this.

tianon:
https://research.swtch.com/vgo-intro is a great article rsc's put out today that's related to this :+1:

AndrienkoAleksandr:
Any news?

tianon:
Yep, `go mod` is official, included in the standard distribution, and supercedes `dep`.

PR #312, opened 2019-12-19
add unzip package

gattytto:
unzip package is needed to decompress protobuf compiler protoc, gunzip (included in the debian image) throws "unknown suffix" error.

required for gRPC .proto compiler release installation (releases are delivered in .zip form).

### DIFF
--- a/1.13/buster/Dockerfile
+++ b/1.13/buster/Dockerfile
@@ -7,6 +7,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 		libc6-dev \
 		make \
 		pkg-config \
+		unzip \
 	&& rm -rf /var/lib/apt/lists/*
 
 ENV GOLANG_VERSION 1.13.5


tianon:
I understand the sentiment, but I'm failing to understand what makes this dependency any different than a library requiring some random C header for `cgo` (which we definitely don't include) -- is there a Go core feature that invokes the `unzip` command directly?

(Given that https://golang.org/pkg/archive/zip/ exists, I'm thinking there probably is not.)

gattytto:
@tianon as for the current time there's no way to "go get" protoc binaries and they are released in a zip form. eclipse/che uses this golang base image without root privileges so one can either build protoc from source or wget the zip release and unpack it. I figured it'd be a "win-win" to just add a basic package like unzip to the golang container at least until there's a "go get" way to obtain protoc protobuf compiler.

tianon:
I get that `protoc` is popular, and I sympathize, but I still don't think I see what makes that project needing some extra dependency any more important from the standpoint of an "upstream-focused Go image" than any other library or tool dependency?

See also #204 where we discussed adding the `dep` tool to the image (and it was ultimately not included for similar reasons).

Given there is not a standard core feature of Go that requires `unzip` and it is pretty trivial to install in a dependent `Dockerfile`, I'm going to close.

PR #319, opened 2020-02-07
Add "stretch" variant of 1.14-rc

### LINKED ISSUE #318: request to add a stretch build for go1.14rc1
Similar to #296. 

### Context
The Go 1.14rc1 image was [recently added](06bc0446f48dfda05b3548bd1f64eb108471fc6b) which is great, but it was only added with a Debian `buster` variant. While `buster` is the current `stable` release, it only moved to `stable` on July 6th 2019, with `stretch` moving to [oldstable](https://wiki.debian.org/DebianOldStable) status. `oldstable` releases like `stretch` will continue to receive security updates for one year ([ref](https://www.debian.org/security/faq#lifespan)) which means they're still a very valid Debian release to use up until July 2020.

### Problem
It can be easier for many applications to be updated with new versions of Go than to update versions of Debian. It will take time for everyone to update applications and their dependencies to be compatible with `buster`. During that period an official Go 1.14 isn't available to folks who are not yet ready to upgrade from `stretch`, but could be running Go 1.14 on production workloads. 

Speaking specifically to my own situation, I'm attempting to use Go 1.14 for stellar/go, but I'm still working through a move to `buster`.

### Proposal / Request
While the `stretch` debian release is still supported with security updates could we support a Go 1.14 image on `stretch` to give folks more time to transition to `buster` while being able to adopt Go 1.14 and run it in development and production workloads?

tianon:
Closes #318

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ matrix:
     - os: linux
       env: VERSION=1.14-rc VARIANT=buster
     - os: linux
+      env: VERSION=1.14-rc VARIANT=stretch
+    - os: linux
       env: VERSION=1.14-rc VARIANT=alpine3.11
     - os: linux
       env: VERSION=1.13 VARIANT=buster
--- a/1.14-rc/stretch/Dockerfile
+++ b/1.14-rc/stretch/Dockerfile
@@ -1,0 +1,50 @@
+FROM buildpack-deps:stretch-scm
+
+# gcc for cgo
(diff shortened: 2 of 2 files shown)


leighmcculloch:
@tianon Thanks for adding this tag! Would it be possible to merge the docker-library-bot/official-images@fdb5edd0d6f7dc053343fcf38eaaebfba591bf15 change upstream to take this to the official images repo?

Issue #328, opened 2020-05-17
Go v1.12 alpine images don't have git installed


dmlambea:
Trying to build a module-enabled project with  `go build -o sample ./cmd/sample/...` fails with an error:

`go: github.com/stretchr/testify@v1.5.1: git init --bare in /go/pkg/mod/cache/vcs/ed2f58bca3966d01dc4666baa48276a4fab360938a8d941050d58e371e2bba77: exec: "git": executable file not found in $PATH
`

More recent versions of Go work flawlessly.

wglambert:
https://github.com/docker-library/golang/issues/250#issuecomment-438486997

https://github.com/docker-library/golang/issues/250#issuecomment-451201761
>As discussed above, adding `git` to the Alpine variant can currently be accomplished with a very trivial `Dockerfile` similar to the following:

```dockerfile
FROM golang:1.11-alpine3.8
RUN apk add --no-cache git
```

Issue #241, opened 2018-09-26
"go get" is broken in alpine image


avoskoboynik:
This makes alpine image useless for many tasks, including compiling/building.

Example:
```
alex ~> docker run -it --rm golang:1.11-alpine go get github.com/aws/aws-sdk-go
go: missing Git command. See https://golang.org/s/gogetcmd
package github.com/aws/aws-sdk-go: exec: "git": executable file not found in $PATH
alex ~> 
```

Same command works fine with standard image:
```
alex ~> docker run -it --rm golang:1.11 go get github.com/aws/aws-sdk-go
alex ~>                                                                                                                                                                                         
```

tianon:
Duplicate of #119, #140, #153, #157, #166, #198, #209, #230, #239.

Issue #230, opened 2018-07-31
cgo runtime error with golang:1.11beta2-alpine3.7


mfridman:
Trying to run a few tests inside this container and seeing cgo errors:

```sh
GOGC=off go test -run ...

# runtime/cgo
exec: "gcc": executable file not found in $PATH
```

Is this by design, where the consumer of the image must `apk add gcc` and fix any dependency issues, `apk add musl-dev` ?

EDIT: the docker image already has those items

https://github.com/docker-library/golang/blob/9d80ad8bdf0abe15bdd3fbc77f92b17e69fe9284/1.11-rc/alpine3.7/Dockerfile#L16-L23

Noticed a similar issue here, golang/go/issues/26307

tianon:
If you look a little further down, the Alpine image removes those again:

https://github.com/docker-library/golang/blob/9d80ad8bdf0abe15bdd3fbc77f92b17e69fe9284/1.11-rc/alpine3.7/Dockerfile#L55

See also https://hub.docker.com/_/golang/, under "`golang:alpine`":

> To minimize image size, it's uncommon for additional related tools (such as `git` or `bash`) to be included in Alpine-based images. Using this image as a base, add the things you need in your own Dockerfile (see the [`alpine` image description](https://hub.docker.com/_/alpine/) for examples of how to install packages if you are unfamiliar).

And https://github.com/docker-library/golang/issues/119.

Issue #209, opened 2018-02-25
git executable not found in golang:alpine image


MatthewEdge:
Attempting to run `go get` in the current golang:alpine image to grab the `dep` command and running into the following error:

```
github.com/golang/dep (download)
go: missing Git command. See https://golang.org/s/gogetcmd
package github.com/golang/dep/cmd/dep: exec: "git": executable file not found in $PATH
The command '/bin/sh -c go get -d -v github.com/golang/dep/cmd/dep' returned a non-zero code: 1
```

Issue persists after complete container and image wipe.

Dockerfile to reproduce (for me at least :D ) :

```
FROM golang:alpine

WORKDIR /go/src/github.com/MatthewEdge/web
COPY . .

# Dependency install
RUN go get -u github.com/golang/dep/cmd/dep
RUN $GOPATH/bin/dep ensure -vendor-only

RUN go install

CMD ["web"]
```

The application being run is the hello world HTTP server for Go. I'm currently getting around this by doing an `apk update && apk install git` at the top of this file but it seemed odd that this image wouldn't have git installed by default?

Thoughts?

Thanks!

tianon:
Duplicate of #157 (see also #119).

In short, `git` is not included in the Alpine variants on purpose (as described in #119 and the image description).

> ## `golang:alpine`
> 
> This image is based on the popular [Alpine Linux project](http://alpinelinux.org), available in [the `alpine` official image](https://hub.docker.com/_/alpine). Alpine Linux is much smaller than most distribution base images (~5MB), and thus leads to much slimmer images in general.
> 
> This variant is highly recommended when final image size being as small as possible is desired. The main caveat to note is that it does use [musl libc](http://www.musl-libc.org) instead of [glibc and friends](http://www.etalabs.net/compare_libcs.html), so certain software might run into issues depending on the depth of their libc requirements. However, most software doesn't have an issue with this, so this variant is usually a very safe choice. See [this Hacker News comment thread](https://news.ycombinator.com/item?id=10782897) for more discussion of the issues that might arise and some pro/con comparisons of using Alpine-based images.
> 
> To minimize image size, it's uncommon for additional related tools (such as `git` or `bash`) to be included in Alpine-based images. Using this image as a base, add the things you need in your own Dockerfile (see the [`alpine` image description](https://hub.docker.com/_/alpine/) for examples of how to install packages if you are unfamiliar).

sethyates:
[...]

PR #345, opened 2020-10-27
Adds Git

Starttoaster:
It appears the upstream alpine image stopped baking git in their image which is essential for `go get`. This PR simply re-adds functionality that was lost from upstream alpine.

### DIFF
--- a/1.14/alpine3.11/Dockerfile
+++ b/1.14/alpine3.11/Dockerfile
@@ -24,6 +24,7 @@ RUN set -eux; \
 		gcc \
 		gnupg \
 		go \
+		git \
 		musl-dev \
 		openssl \
 	; \
--- a/1.14/alpine3.12/Dockerfile
+++ b/1.14/alpine3.12/Dockerfile
@@ -24,6 +24,7 @@ RUN set -eux; \
 		gcc \
 		gnupg \
 		go \
+		git \
 		musl-dev \
 		openssl \
 	; \
--- a/1.15/alpine3.12/Dockerfile
+++ b/1.15/alpine3.12/Dockerfile
@@ -24,6 +24,7 @@ RUN set -eux; \
 		gcc \
 		gnupg \
 		go \
+		git \
 		musl-dev \
 		openssl \
 	; \
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -18,7 +18,7 @@ RUN set -eux; \
 		gcc \
 		gnupg \
 		go \
-                git \
+		git \
 		musl-dev \
 		openssl \
 	; \
--- a/Dockerfile-alpine.template
+++ b/Dockerfile-alpine.template
@@ -18,6 +18,7 @@ RUN set -eux; \
 		gcc \
 		gnupg \
 		go \
+                git \
 		musl-dev \
 		openssl \
 	; \


tianon:
I'm not sure where you're seeing that Alpine used to include `git` by default?  Our `golang:alpine` image has never included `git`, and that's intentional (https://github.com/docker-library/golang/issues/250#issuecomment-438486997; #119, #140, #153, #157, #166, #198, #209, #230, #239, #241, #251, #328).

Starttoaster:
> I'm not sure where you're seeing that Alpine used to include `git` by default? Our `golang:alpine` image has never included `git`, and that's intentional ([#250 (comment)](https://github.com/docker-library/golang/issues/250#issuecomment-438486997); #119, #140, #153, #157, #166, #198, #209, #230, #239, #241, #251, #328).

Huh, interesting. We use `golang:alpine` in a lot of our CI jobs and we only very recently began noticing them failing due to git not being installed. Using the exact same CI templates from other projects that also were not installing git, their `go get`s were succeeding. But I guess you would be an authoritative source to say I'm imagining that. :smile:

tianon:
Are the working projects using `go.mod`?  I believe it has the ability to download specific versions from proxy.golang.org without `git` installed.

klmitch:
> Are the working projects using `go.mod`? I believe it has the ability to download specific versions from proxy.golang.org without `git` installed.

Yes, they are; we're exclusively using `go.mod`-style code.  One of our failing CI jobs is actually executing just `go vet` and is encountering the following error:
```
go vet -mod=readonly ./...
go: gitlab.com/tmobile/cdp/libraries/cdp-error@v0.1.0: git init --bare in /go/pkg/mod/cache/vcs/5284ea28f371dd764ab19f2614118ef1a885457d03dbd84605c2c2d10e430d11: exec: "git": executable file not found in $PATH
```

tianon:
Yeah, that sounds like it's probably the case!  For running `go vet`, you probably want to use non-Alpine images generally, but in this specific case I think you can run your own instance of `proxy.golang.org` which could then include your internal dependencies? (Haven't looking into the details of that myself)

Starttoaster:
> Are the working projects using `go.mod`? I believe it has the ability to download specific versions from proxy.golang.org without `git` installed.

That they are. Hm, I'll have to play with this a bit to confirm. But this still strikes me as a basic functionality add of the Go language. I mean, if the argument is that we can just install git in our own 2 liner Dockerfile that FROM's this image because adding git is undue burden from this project's side of things (or from inside the CI job), then couldn't the argument be made that installing Go is also undue burden? :smile: Not trying to come across as crass, just trying to understand the philosophy there.

Starttoaster:
Also, sorry for creating a dupe PR :smile:

klmitch:
We tend to prefer using alpine-based images for our CI, because they _are_ smaller, and thus load faster.  We've not run into this particular issue before because we don't have very much go code within the company yet, but the image not having a known dependency of go definitely violated the principle of least surprise for me.  We have ways of working around this internally--basically, adding `apk add git`--, but that workaround is also going to slow down our CI a little.  We _could_, I suppose, build our own variant of the upstream `golang` image, but we really don't want to try to maintain such images ourselves, which is why we haven't done that and indeed have resisted requests to do that from our customers.  (It might be worth mentioning that we're from the CI team of a very large company...)

tianon:
The point is not "undue burden" but rather "unnecessary size" (which then *cannot* be reclaimed by image users).

I think https://github.com/docker-library/golang/issues/250#issuecomment-451201761 phrases this better than I could rephrase again. :sweat_smile:

(That note about Alpine being entirely unsupported/unmaintained in the Go ecosystem has also since been documented in our image description via https://github.com/docker-library/docs/pull/1749)

klmitch:
> The point is not "undue burden" but rather "unnecessary size" (which then _cannot_ be reclaimed by image users).

True; what this means is that you prioritize what is necessary, which you have done.  I'm simply arguing that git is necessary in some cases :)

> (That note about Alpine being entirely unsupported/unmaintained in the Go ecosystem has also since been documented in our image description via [docker-library/docs#1749](https://github.com/docker-library/docs/pull/1749))

Interesting.  I do have to point out that that comment is not accurate; the alpine image _does_ indeed include both bash and gcc; it's just missing git.

Starttoaster:
> The point is not "undue burden" but rather "unnecessary size" (which then _cannot_ be reclaimed by image users).
> 
> I think [#250 (comment)](https://github.com/docker-library/golang/issues/250#issuecomment-451201761) phrases this better than I could rephrase again. sweat_smile
> 
> (That note about Alpine being entirely unsupported/unmaintained in the Go ecosystem has also since been documented in our image description via [docker-library/docs#1749](https://github.com/docker-library/docs/pull/1749))

That's fair.. FTR, I still think this is a pretty simple value add, even if it adds a few MB to the image. But I will defer to your opinion that since git may not under every circumstance be necessary (even if IMO common) then it should not be added.

Starttoaster:
@tianon just wanted to revist this PR since I see it's still open. Wanted to make the point that for the debian variants of this image, the template in this repo is specifying `FROM buildpack-deps:{{ env.variant }}-scm`. The `scm` suffix to that tag implies that it's recognized that git is a basic and assumed addition to any Docker image containing Go. 

Looking at the documentation for that buildpack-deps image: https://hub.docker.com/_/buildpack-deps

```
This variant is based on curl, but also adds various source control management tools. As of this writing, the current list of included tools is bzr, git, hg, and svn.
```

Pulling some sample image tags to my local, I can see that `golang:1` (debian based) is 839MB whereas `golang:1-alpine` is 299MB. This rises to 317MB if you add git, but it is still much more lightweight than the debian variants. Noting that your debian variants include git, as well as several other SCM tools like svn, I'm not sure why the position is still being held that git should not be in the Alpine based golang image? Seems to me like your images set the precedent that Debian _requires_ git, where Alpine doesn't...because?

tianon:
The default (Debian) image tags target convenience -- see, for example, `python:3.8` which is `FROM buildpack-deps:buster` vs `python:3.8-slim` which is instead `FROM debian:buster-slim`.  If we were to add "slim" variants of the Debian images here, they would also not include Git.

In contrast, the Alpine image tags almost always target size (because that's typically the main reason folks choose Alpine in the first place), and for *many* uses of Go, `git` is definitely not required (even _less_ so when using Go Modules with a properly configured `GOPROXY`).

(In other words, I definitely still stand by https://github.com/docker-library/golang/issues/250#issuecomment-451201761.)

Starttoaster:
> In contrast, the Alpine image tags almost always target size

This is a fair point. Though there's something like a 500MB difference between the alpine and debian image variants... Asking someone that needs git to have their CI download 500 extra MB because they want git is a pretty tall order. On the flipside, installing git for every CI job that needs it is also tedious and a tall order in its own way.

To counterpoint, is it _always_ necessary for a golang image to contain ca-certificates and openssl? Bash?

tianon:
> To counterpoint, is it _always_ necessary for a golang image to contain ca-certificates and openssl? Bash?

`openssl` and `bash` are only added temporarily for the purposes of the Go build/install process -- they are removed as soon as it's finished.

In the case of `ca-certificates`, it's definitely necessary in a *much* higher percentage of use cases than `git` is (since it'll be used any time Go talks to `GOPROXY` or any time an application reaches out to the internet), _and_ is only ~512kB (although technically, most of it is already included in `alpine:3.12` itself, so we could probably do some testing around that and remove that too, but I don't know that the benefit in doing so is strong and I certainly won't be exploring it).

tianon:
Also, we're suggesting that users who want a *supported* installation of Go download those additional 500MB, not just folks who want `git` (because Go on Alpine is explicitly unsupported to the point that the upstream Go developers disabled the buildbot workers they had testing it because nobody was working on fixing it so it was just permanently failing).

balarajupetta (approved):

yosifkit:
Closing old issue; since our goal for `alpine` based images is small size, we do not want to add `git` to it.

This is documented on the [Docker Hub description](https://github.com/docker-library/docs/tree/22560eae6395a0f1e51ee7d180dcd1077ab3202c/golang#golangversion-alpine) about the `alpine` variants:

>To minimize image size, additional related tools (such as `git`, `gcc`, or `bash`) are not included in Alpine-based images. Using this image as a base, add the things you need in your own Dockerfile (see the [`alpine` image description](https://hub.docker.com/_/alpine/) for examples of how to install packages if you are unfamiliar). See also [docker-library/golang#250 (comment)](https://github.com/docker-library/golang/issues/250#issuecomment-451201761) for a longer explanation.

PR #361, opened 2021-03-03
Add "stretch" variants

### LINKED ISSUE #344: request to add a stretch build for go1.15
Similar to #296 and #318. 

### Context
The Go 1.15 image was recently added which is great, but it was only added with a Debian `buster` variant. While `buster` is the current `stable` release, it only moved to `stable` on July 6th 2019, with `stretch` being in LTS status until 2022.

### Problem
It can be easier for many applications to be updated with new versions of Go than to update versions of Debian. It will take time for everyone to update applications and their dependencies to be compatible with `buster`. During that period an official Go 1.15 isn't available to folks who are not yet ready to upgrade from `stretch`, but could be running Go 1.15 on production workloads. 

Speaking specifically to my own situation, I'm attempting to use Go 1.15 for stellar/go, but I'm still working through a move to `buster`.

### Proposal / Request
While the `stretch` debian release is still supported with security updates could we support a Go 1.15 image on `stretch` to give folks more time to transition to `buster` while being able to adopt Go 1.15 and run it in development and production workloads?

### LINKED ISSUE #351: request to add a stretch build for go1.16-beta1 and go1.16
Similar to #296, #318, and #344.

### Context
The Go 1.16-beta1 image was recently added which is great, but it was only added with a Debian `buster` variant. While `buster` is the current `stable` release, it only moved to `stable` on July 6th 2019, with `stretch` being in LTS status until 2022.

### Problem
It can be easier for many applications to be updated with new versions of Go than to update versions of Debian. It will take time for everyone to update applications and their dependencies to be compatible with `buster`. During that period an official Go 1.16 isn't available to folks who are not yet ready to upgrade from `stretch`, but could be running Go 1.16 on production workloads. 

Speaking specifically to my own situation, I'm attempting to use Go 1.16 for stellar/go, but I'm still working through a move to `buster`.

Go 1.15 images also don't provide a stretch variant (#344) and this means once Go 1.16 is released there will be no version of Go in an official Docker image based on `stretch` that is receiving security updates since only the last two releases of Go receive security patches.

### Proposal / Request
While the `stretch` debian release is still supported with security updates could we support a Go 1.16 image on `stretch` to give folks more time to transition to `buster` while being able to adopt Go 1.16 and run it in development and production workloads?

tianon:
Given the glibc ABI break between buster and stretch, it is thus inordinately difficult to build binaries that work on older distributions (binaries build on buster won't run on stretch, for example).

~~This also re-enables arm32v5 for the stretch images as we now set `GOARM` and friends explicitly for build scenarios.~~

Closes #344
Closes #351

### DIFF
--- a/1.15/stretch/Dockerfile
+++ b/1.15/stretch/Dockerfile
@@ -1,0 +1,123 @@
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "apply-templates.sh"
+#
+# PLEASE DO NOT EDIT IT DIRECTLY.
+#
+
+FROM buildpack-deps:stretch-scm
+
+# gcc for cgo
+RUN apt-get update && apt-get install -y --no-install-recommends \
+		g++ \
+		gcc \
+		libc6-dev \
+		make \
+		pkg-config \
+	&& rm -rf /var/lib/apt/lists/*
+
+ENV PATH /usr/local/go/bin:$PATH
+
+ENV GOLANG_VERSION 1.15.8
+
+RUN set -eux; \
+	\
+	dpkgArch="$(dpkg --print-architecture)"; \
+	url=; \
+	case "${dpkgArch##*-}" in \
+		'amd64') \
+			url='https://storage.googleapis.com/golang/go1.15.8.linux-amd64.tar.gz'; \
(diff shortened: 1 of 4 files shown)


tianon:
> ~~This also re-enables arm32v5 for the stretch images as we now set `GOARM` and friends explicitly for build scenarios.~~

I had to back out this bit because after testing it still fails to build. :disappointed:

PR #367, opened 2021-04-22
changed chmod privs to be more secure

eric-w-hart:
777 is a known risk.
755 is a better alternative


eric-w-hart:
Closing per call out by @tianon

PR #371, opened 2021-06-10
Add 1.17-rc

tianon:


Dockerfile-alpine.template
```diff
@@ -89,13 +89,13 @@ RUN set -eux; \
...
 	\
-	apk del --no-network .build-deps; \
-	\
 # pre-compile the standard library, just like the official binary release tarballs do
 	go install std; \
 # go install: -race is only supported on linux/amd64, linux/ppc64le, linux/arm64, freebsd/amd64, netbsd/amd64, darwin/amd64 and windows/amd64
 #	go install -race std; \
 	\
+	apk del --no-network .build-deps; \
+	\
```
tianon, line 98: For some reason, `go install std` in 1.17 needs `gcc` (where previous versions did not), hence this change.  I'd love to find a good reference for why. :sweat_smile:

PR #381, opened 2021-08-16
Update to 1.17 (GA)

### LINKED ISSUE #356: Add Debian 11 base image
The latest version 11 of Debian version 11 (bullseye) is released. So can we have a golang docker image with debian 11 as base.

https://release.debian.org/bullseye/freeze_policy.html

The current images of golang have openldap and bash vulnerabilities due to the underlying debian

Debian Security   Update for openldap (DSA 4792-1) | CVE-2020-25709,CVE-2020-25710
GNU Bash Privilege Escalation   Vulnerability for Debian (Zero Day) | CVE-2019-18276
The current golang images have 4.1 version of bash that has this vulnerability, upgrading to debian 11 will provide 5.1 bash which will have the issue fix

tianon:
Closes #356


tianon:
I've updated this to include Bullseye variants, but I've also left Stretch because the variants here don't build from source (except on architectures that are esoteric enough that upstream doesn't provide builds for them), so the load is low.

yosifkit (approved):

yosifkit:
https://github.com/docker-library/official-images/pull/10739#issuecomment-899916032

PR #385, opened 2021-08-20
Combine Linux templates into one

tianon:
This helps ensure we have even better consistency between them than before.

I could've used more conditionals here to reduce the amount of change, but it didn't seem worthwhile when it's pretty harmless to have them more consistent with one another instead.

(Easiest to review without whitespace)


Dockerfile-linux.template
```diff
@@ -0,0 +1,201 @@
...
+				export GOROOT_BOOTSTRAP="/usr/local/go-$GOOS-$GOARCH-bootstrap"; \
+				"$GOROOT_BOOTSTRAP/bin/go" version; \
+			fi; \
+{{ ) else "" end -}}
+			./make.bash; \
+{{ if need_386_hack then ( -}}
+			if [ -n "${GO386:-}" ]; then \
+				rm -rf "$GOROOT_BOOTSTRAP"; \
+			fi; \
+{{ ) else "" end -}}
```
yosifkit, line 164: This looks silly to make a function to use twice to avoid having an else that duplicates the `./make.bash` line. (It's not new to this PR, so we can fix it later).

PR #395, opened 2021-10-09
Finally remove Johan (❤)

tianon:
FYI @proppy :joy: :heart:

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -72,8 +72,7 @@ cat <<-EOH
 # this file is generated via https://github.com/docker-library/golang/blob/$(fileCommit "$self")/$self
 
 Maintainers: Tianon Gravi <admwiggin@gmail.com> (@tianon),
-             Joseph Ferguson <yosifkit@gmail.com> (@yosifkit),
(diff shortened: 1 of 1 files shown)


tianon:
(The main motivation here is that https://github.com/docker-library/official-images/pull/10879 will start pinging you on every image update, so we want to save you from that pain :sweat_smile:)

proppy:
LGTM and thanks :)

Sorry, I wasn't more helpful "lately" (for years) let me know if there is any way I can contribute!

tianon:
(Also, if you don't care about the notification spam, we can just close this too, and keep having smiles together! :joy:)

proppy:
There are a few other places that bring "smiles":
https://github.com/search?q=org%3Adocker-library+proppy&type=code

Do you want me to clean them up as well? (maybe that one way I can responsibly contribute if I'm not involved in the project anymore).

tianon:
If you want to, sure!  However, none of those are going to be filling up your inbox (and don't bother us at all), so it's up to you. :smile:

So I guess the verdict is that you'd rather not have the inbox spam and we should merge this? :sweat_smile:

PR #405, opened 2021-12-22
feat: Add windowsservercore-20H2 variant 

afiune:
Adds a new variant to generate docker images for `windows/servercore:20H2`

tenor-154511983


tianon:
Thanks for the contribution!  Can you elaborate a little bit on the use case for this version that's just slightly older than `ltsc2022`?

afiune:
@tianon I tried to use `ltsc2022` but my CI pipeline was complaning with:

Screen Shot 2021-12-22 at 1 12 57 PM

I am not a Windows expert but since we are running on Windows 10 (20H2) I thought adding that version would be a good idea, I have built the image (pretty easy and super straightforward so kuddos to y'all for making it super simple) and publish it at https://hub.docker.com/r/lacework/win-golang for now, it works just fine so I wanted to contribute back.

Screen Shot 2021-12-22 at 1 17 02 PM

yosifkit:
As you discovered, you can only run Windows images that are equal to or older than the host kernel. We run our Windows Server builder on LTS releases since they can be upgraded from one to the next and as such cannot build images newer than the most recent LTS. We usually only publish images for LTS releases since by the time we can build any of the semi-annual channel releases, they are either [end of life or close to it](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle). With `20H2` end of life in May 2022, I don't think it makes sense to add them now.

I think some possible workarounds are upgrading the host, using the `windowsservercore-1809` images, building `20H2` golang images for your own use,  or maybe building your go application in a Linux container.

afiune:
Alright, I will close this PR. Thank you for the prompt response!

Issue #416, opened 2022-03-17
Missing `linux/ppc64le` for Golang 1.18 image


jauderho:
Go 1.18 was just released but it appears that the build job for linux/ppc64le have failed and need to be respun.

This is in reference to https://github.com/docker-library/official-images/pull/12045#issuecomment-1068723763 

Per @DasSkelett,

* https://doi-janky.infosiftr.net/job/multiarch/view/images/view/golang/job/ppc64le/job/golang/
* https://doi-janky.infosiftr.net/job/multiarch/view/images/view/golang/job/ppc64le/job/golang/587/console

FWIW, it looks like PR docker-library/official-images#12055 is inbound so it's unclear if we want to wait until that rolls out.

tianon:
The failure looks like a pretty consistent failure in either `go` from Alpine itself (installed fresh within that layer, so https://github.com/docker-library/official-images/pull/12055 is not likely to cause any change), or in the bootstrap version of `go` that gets built from Alpine's `go`:

```console
...
+ ./make.bash
Building Go cmd/dist using /usr/lib/go. (go1.16.10 linux/ppc64le)
Building Go toolchain1 using /usr/lib/go.
Building Go bootstrap cmd/go (go_bootstrap) using Go toolchain1.
Building Go toolchain2 using go_bootstrap and Go toolchain1.
Building Go toolchain3 using go_bootstrap and Go toolchain2.
Building packages and commands for linux/ppc64le.
go tool dist: FAILED: /usr/local/go/bin/go list -gcflags=all= -ldflags=all= -f={{if .Stale}}	STALE {{.ImportPath}}: {{.StaleReason}}{{end}} std cmd: signal: segmentation fault (core dumped)
```

DasSkelett:
There's now an upstream Go issue tracking this, the Alpine people have hit the same trying to build the package for their repository:
https://github.com/golang/go/issues/51787
On EOF their build logs: https://gitlab.alpinelinux.org/alpine/aports/-/jobs/661598#L1073

jauderho:
Looks like ppc64le builds are showing up now BUT only for certain tags. ppc64le builds are available for `alpine` and `alpine3.15` but not available for `1.18.0-alpine` or `1.18.0-alpine3.15`

See https://hub.docker.com/_/golang?tab=tags&page=1&name=alpine

yosifkit:
[...]

PR #420, opened 2022-04-13
Revert upstream-applied ppc64le patch

DasSkelett:
The build for ppc64le [fails currently](https://doi-janky.infosiftr.net/job/multiarch/view/images/view/golang/job/ppc64le/job/golang/), due to [not being able to apply the patch](https://doi-janky.infosiftr.net/job/multiarch/view/images/view/golang/job/ppc64le/job/golang/601/execution/node/101/log/) from #417 for #416, as it is already part of Go 1.18.1:
* https://github.com/golang/go/issues/51874
* https://github.com/golang/go/commit/0bf8319883298dbeea81444cb704d8c0e9935bae

Reverts #417

### DIFF
--- a/1.18/alpine3.14/Dockerfile
+++ b/1.18/alpine3.14/Dockerfile
@@ -78,14 +78,6 @@ RUN set -eux; \
 			go \
 			musl-dev \
 		; \
-		if [ "$GOARCH" = 'ppc64le' ]; then \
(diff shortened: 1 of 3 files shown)


GoliathLabs:
Yeah, would be nice if this PR get's merged as we're currently unable to build for ppc64le

tianon:
Oops, great catch, thank you!

PR #421, opened 2022-04-28
Set GOAMD64 explicitly to v1

tianon:
This is already the default, but it seems like a good idea to be explicit about it.

(See also https://tip.golang.org/doc/go1.18#amd64)


tianon:
(Not that it's super important, but this also matches Debian's `amd64` baseline: https://wiki.debian.org/ArchitectureSpecificsMemo#Architecture_baselines)

yosifkit (approved):

PR #425, opened 2022-05-24
support alpine 3.16 & remove alpine 3.14

techknowlogick:


achton:
Why remove 3.14 though? It [has official support until May 2023](https://alpinelinux.org/releases/)?

tianon:
Our policy for Alpine releases for this image is to only support the latest two releases, especially since they have to build from source (as they're not an officially supported distribution for Go upstream).

PR #441, opened 2022-11-14
Ensure nsswitch.conf config for netgo

yosifkit:
Alpine 3.16 has updated to [include an `nsswitch.conf`](https://git.alpinelinux.org/aports/commit/main/alpine-baselayout?h=3.16-stable&id=348653a9ba0701e8e968b3344e72313a9ef334e4). But we need to ensure 3.15 still works as well.

```console
Step 1/10 : FROM alpine:3.15
Step 3/10 : RUN set -eux; 	if [ -e /etc/nsswitch.conf ]; then 		grep '^hosts: files dns' /etc/nsswitch.conf; 	else 		echo 'hosts: files dns' > /etc/nsswitch.conf; 	fi
 ---> Running in 79d0f913ddf7
+ '[' -e /etc/nsswitch.conf ]
+ echo 'hosts: files dns'
```
```console
Step 1/10 : FROM alpine:3.16
...
Step 3/10 : RUN set -eux; 	if [ -e /etc/nsswitch.conf ]; then 		grep '^hosts: files dns' /etc/nsswitch.conf; 	else 		echo 'hosts: files dns' > /etc/nsswitch.conf; 	fi
 ---> Running in b9440f91dc03
+ '[' -e /etc/nsswitch.conf ]
+ grep '^hosts: files dns' /etc/nsswitch.conf
```

Related to https://github.com/docker-library/docker/pull/384

### DIFF
--- a/1.18/alpine3.15/Dockerfile
+++ b/1.18/alpine3.15/Dockerfile
@@ -8,10 +8,16 @@ FROM alpine:3.15
 
 RUN apk add --no-cache ca-certificates
 
-# set up nsswitch.conf for Go's "netgo" implementation
-# - https://github.com/golang/go/blob/go1.9.1/src/net/conf.go#L194-L275
+# ensure that nsswitch.conf is set up for Go's "netgo" implementation
+# - https://github.com/golang/go/blob/go1.19.3/src/net/conf.go#L227-L303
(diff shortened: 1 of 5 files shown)


Heviwulansari:
https://www.docker.com/products/docker-desktop

PR #442, opened 2022-11-21
Add support for alpine 3.17, remove alpine 3.15

squizzi:
This PR adds support for alpine 3.17 and removes support for 3.15 to adhere to the policy of only supporting two major releases.

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -9,17 +9,6 @@
 FROM alpine:{{ alpine_version }}
 
 RUN apk add --no-cache ca-certificates
-
-# ensure that nsswitch.conf is set up for Go's "netgo" implementation
-# - https://github.com/golang/go/blob/go1.19.3/src/net/conf.go#L227-L303
-# - docker run --rm debian grep '^hosts:' /etc/nsswitch.conf
-# Alpine 3.16 includes nsswitch.conf
-RUN set -eux; \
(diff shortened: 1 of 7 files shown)


squizzi:
Apparently alpine:3.17 is still in RC stages so I jumped the gun a bit on this, but it should be out officially by end of week.

tianon:
Nice, thanks! :+1:

We should also ditch this entire `RUN` line (and associated comments) now that we don't need to deal with 3.15 anymore: :eyes:

https://github.com/docker-library/golang/blob/326acd5eed36954174ba8b3b6d0efda96087e18a/Dockerfile-linux.template#L13-L22

(I'm happy to take over and make this change if you would rather not do so!)

squizzi:
> We should also ditch this entire RUN line (and associated comments) now that we don't need to deal with 3.15 anymore: eyes

How's that look?

Dockerfile-linux.template
```diff
@@ -10,16 +10,6 @@ FROM alpine:{{ alpine_version }}
 
 RUN apk add --no-cache ca-certificates
```
tianon, line 12: Off-by-one; this empty line should go too :sweat_smile: 

```suggestion
```

(Again, happy to take care of this nit for you if you'd prefer!)

squizzi: :cry:

tianon:
Nice thank you!  (and thanks for sticking with it!)

PR #456, opened 2023-04-05
Add debian:bookworm distro for 1.20 and 1.19

sitano:
There are many CVEs not-fixed in bullseye that we need a newer version that is better with critical and high issues.

### DIFF
--- a/1.19/bookworm/Dockerfile
+++ b/1.19/bookworm/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM buildpack-deps:buster-scm
+FROM buildpack-deps:bookworm-scm
 
 # install cgo-related dependencies
 RUN set -eux; \
--- a/1.20/bookworm/Dockerfile
+++ b/1.20/bookworm/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM buildpack-deps:buster-scm
+FROM buildpack-deps:bookworm-scm
 
 # install cgo-related dependencies
 RUN set -eux; \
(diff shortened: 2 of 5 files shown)


tianon:
Do you have some examples?  I built this locally so I could compare, and I'm coming up with a different result:

```console
$ docker scout compare golang:bookworm --to golang:bullseye
WARN 'docker scout compare' is in early preview and its behaviour might change in the future
    v SBOM of image already cached, 300 packages indexed
    v SBOM of image already cached, 300 packages indexed


  ## Overview

                      │           Analyzed Image           │          Comparison Image
  ────────────────────┼────────────────────────────────────┼─────────────────────────────────────
    Image reference   │  golang:bookworm                   │  golang:bullseye
      vulnerabilities │    0C     1H     3M    35L     5?  │    0C     0H     0M    60L     1?
                      │           +1     +3    -25     +4  │
                      │                                    │
    Base image        │  buildpack-deps:bookworm-scm       │  buildpack-deps:bullseye-scm
      tags            │ also known as                      │ also known as
                      │   • testing-scm                    │   • scm
                      │                                    │   • stable-scm
      vulnerabilities │    0C     0H     1M    28L     8?  │    0C     0H     0M    41L     1?

...
```

There is some additional context in https://github.com/docker-library/faq#why-does-my-security-scanner-show-that-an-image-has-cves specifically about CVEs.  The Debian Security Team is _extremely_ proactive about important security fixes, especially in the active `stable` release (which is what Bullseye is).

(The single "high" that comes up for me on Bookworm is https://dso.docker.com/cve/CVE-2023-0464, which from https://security-tracker.debian.org/tracker/CVE-2023-0464 you can see is marked as essentially "wontfix" by the Debian Security Team because it's a "minor issue": `[bullseye] - openssl <no-dsa> (Minor issue)`)

Once Bookworm is officially released, we will definitely update this repository to support it, but until then we plan to continue to wait (especially as Debian Bookworm is not actively supported by the Debian Security Team until it is officially released).

sitano:
@tianon the primary image for me was ruby. putting everything into bookworm I had to build golang programs under it as well.

I use `trivy` for scans.  Here it is its results for `golang:1.20-bullseye`:

```
$ docker run -v /var/run/docker.sock:/var/run/docker.sock aquasec/trivy image golang:1.20-bullseye

...

golang:1.20-bullseye (debian 11.6)
==================================
Total: 474 (UNKNOWN: 0, LOW: 305, MEDIUM: 76, HIGH: 85, CRITICAL: 8)

...
```

Please, see the attachments for full report.

Maybe those critical and high CVEs are not important but I needed to have my Harbor scans clean. Moving to bookworm was the simplest option. I know for example that at least 1 critical CVE there is not that important `[CVE-2019-8457](https://avd.aquasec.com/nvd/cve-2019-8457) Critical libdb5.3 5.3.28+dfsg1-0.8` but I don't want to manually check every one of them. 

For some of those issues debian devs reported that it will be too complicated to either migrated or mitigate affected versions. It's hard to manually ensure hundreds of those are actually safe...
[golang-1.20-bullseye-trivy-report.log](https://github.com/docker-library/golang/files/11169800/golang-1.20-bullseye-trivy-report.log)

yosifkit:
Rebased on master. Added `bookworm` to `versions.sh`. Dropped Debian `buster`. Applied templates.

Issue #472, opened 2023-06-27
proposal: set GOTOOLCHAIN=local (or =path) in our image


tianon:
Per https://go.dev/doc/toolchain, it might make sense for us to set `GOTOOLCHAIN=local` (or even `GOTOOLCHAIN=path`) by default in our images.  I'll follow this up with a comment describing more long-form why I think this might be appropriate here, but wanted to get an issue open where it can be discussed properly.

tianon:
@rsc I really hope you don't mind being pinged for your opinion (or help directing to the right folks for an opinion) on how we can best represent Go here; I've tried to make sure this has enough information for you help guide us without having to be totally well versed in how Go is commonly used in containers (just in case you aren't) :heart: :sweat_smile: :bow:

One of our highest goals in maintaining this image is to appropriately represent Go upstream in the way we believe they might represent themselves (https://github.com/docker-library/official-images/tree/78bad2f7ead1688af35d768fa1a23939fff77f89#what-are-official-images, bullet point 5).  We're obviously not perfect at this (or balancing this with end user expectations/desires), but we do try very hard to do so unobtrusively.

This has come up most recently in the context of `GOTOOLCHAIN` (https://go.dev/doc/toolchain), with some users expecting/requesting that we set it explicitly to `local` in our images for 1.21+, and us hesitating to do so because the default was chosen explicitly and intentionally upstream (after some very long debates on the subject).

We do *not* want or intend to reopen those debates (and any unproductive comments attempting to do so here will be hidden as off topic), but we do think the way that users interact with this image currently perhaps _does_ warrant a change in that default behavior.

Typically, users of this image will have a `Dockerfile` which looks something like this:

```dockerfile
FROM golang:X.Y
# or
FROM golang:X.Y.Z
# ...
COPY my-code /some/path
RUN go build ...
```

As you can see, there is usually a very _clear_ user intent that they desire Go version X.Y (or X.Y.Z), and the argument here is that it would be very surprising if a `go.mod` that contains a newer `go X.Y+N` were to not error, but rather download a newer toolchain (again, given the very clear user intent in the `Dockerfile`).

In short, the proposal is that we _should_ set `GOTOOLCHAIN=local` (or perhaps `GOTOOLCHAIN=path` so that `go get go@xxx` still works appropriately and is used correctly, for users who might be mixing and matching multiple Go versions in a single image), but we want to make sure we're doing so in a way that is respectful to the origins of the default choice. :heart:

[...]

PR #491, opened 2023-09-16
Set `GOTOOLCHAIN` to prevent go auto-upgrade

yosifkit:
Fixes https://github.com/docker-library/golang/issues/472

Using an image that is too old for your project's minimum `go` version will cause `go` to fail and help inform the user that they are not using a new enough `go` for their project. So, rather than having `go` automatically upgrading the toolchain (`GOTOOLCHAIN=auto`, the default), we have opted to move the image to encourage users to upgrade the image to get newer versions of `go`.

(See also the comments on #472)

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -201,6 +201,12 @@ RUN set -eux; \
 {{ ) else "" end -}}
 	go version
 
+{{ if [ "1.20" ] | index(env.version | rtrimstr("-rc")) then "" else ( -}}
+# don't auto-upgrade the gotoolchain
+# https://github.com/docker-library/golang/issues/472
+ENV GOTOOLCHAIN=local
+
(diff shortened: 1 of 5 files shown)


Dockerfile-linux.template
```diff
@@ -201,6 +201,12 @@ RUN set -eux; \
 {{ ) else "" end -}}
 	go version
 
+{{ if [ "1.20" ] | index(env.version | rtrimstr("-rc")) then "" else ( -}}
+# don't auto-upgrade the gotoolchain
+# https://github.com/docker-library/golang/issues/472
+ENV GOTOOLCHAIN=local
```
tianon, line 207: We might want to move this up a bit higher (later) to make sure our `go` invocations during install and verification don't somehow accidentally download. 👀

PR #492, opened 2023-09-20
Install git-lfs to support Go modules with large-files

danielorbach:
This pull-request installs `git-lfs` on the Linux images to support the download of Go modules from private repositories. The following is true even for public repositories that are downloaded directly (as opposed to downloading from a module-proxy).

The Go toolchain uses the local git installation to fetch archives of modules from Git servers directly. My company commits large files using Git LFS which happen to be part of our Go module source tree. Without _git-lfs_ installed, the local git installation does not automatically resolve those files which causes "_checksum mismatch_" against the committed `go.sum`.

@yosifkit I have tested that the generated images in-fact resolve the Issue. I'd appreciate your review.
Looking for your guidance on getting this enhancement to the community,

### DIFF
--- a/1.20/alpine3.17/Dockerfile
+++ b/1.20/alpine3.17/Dockerfile
@@ -108,6 +108,9 @@ RUN set -eux; \
 	\
 	go version
 
+# install git-lfs for Go modules with LFS committed files
+RUN apk add --no-cache git-lfs
+
 ENV GOPATH /go
 ENV PATH $GOPATH/bin:$PATH
 RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 1777 "$GOPATH"
--- a/1.20/alpine3.18/Dockerfile
+++ b/1.20/alpine3.18/Dockerfile
@@ -108,6 +108,9 @@ RUN set -eux; \
 	\
 	go version
 
+# install git-lfs for Go modules with LFS committed files
+RUN apk add --no-cache git-lfs
+
 ENV GOPATH /go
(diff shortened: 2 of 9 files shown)


LaurentGoderre:
We usually explicitly avoid doing this and let downstream images add those. We had similar requests in the Node image for git to support package manager but if it's added in the base image it can't be removed in subsequent layers.

yosifkit:
We appreciate your contribution ❤️🙇 

I agree with @LaurentGoderre that the added size would not acceptable for all users of the images. While useful, `git-lfs` isn't a core feature of `git` but an extension, and as such is not expected to be installed wherever `git` is.

For users that need `git-lfs` it is as simple as `apt-get` (or `apk`) installing it or downloading it from `https://github.com/git-lfs/git-lfs/releases` in their custom image.
```dockerfile
FROM golang:1.21
RUN apt-get update && apt-get install -y git-lfs && rm -rf /var/lib/apt/lists/*
```

Similar issue/request: https://github.com/docker-library/buildpack-deps/issues/112

danielorbach:
@LaurentGoderre @yosifkit Waiting on you instructions to close this pull-request...

danielorbach:
@LaurentGoderre I'd be glad to extend the documentation!

Can you point me to the appropriate repository? I am a bit lost regarding the many repositories involved.

LaurentGoderre:
@yosifkit correct me if im wrong but I think it's this file here: https://github.com/docker-library/docs/blob/master/golang/content.md

PR #497, opened 2023-12-07
Add Alpine 3.19 variant

jnoordsij:
Modelled after #460.

This PR adds a variant for Alpine 3.19 and drops the 3.17 variant at the same time.

See also https://alpinelinux.org/posts/Alpine-3.19.0-released.html.

*~~Note: the Alpine 3.19 base Docker image is yet to appear [here](https://hub.docker.com/_/alpine/tags), so automated checks will fail up till then; see also https://github.com/docker-library/official-images/pull/15852.~~*

*Edit: the Alpine 3.19 base images are now available, so this is now ready for a retrigger of the workflow!*

### DIFF
--- a/versions.sh
+++ b/versions.sh
@@ -148,8 +148,8 @@ for version in "${versions[@]}"; do
 			"bookworm",
 			"bullseye",
 			(
-				"3.18",
-				"3.17"
+				"3.19",
+				"3.18"
 			| "alpine" + .),
 			if .arches | has("windows-amd64") and .["windows-amd64"].url then
 				(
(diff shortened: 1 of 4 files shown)


versions.sh
```diff
@@ -148,8 +148,8 @@ for version in "${versions[@]}"; do
 			"bookworm",
 			"bullseye",
 			(
-				"3.18",
-				"3.17"
+				"3.19",
+				"3.18"
```
tianon, line 152: Might be worth adding our `empty` hack here to make the diff cleaner next time (not worth cycling this time IMO :heart:).

tianon:
Merging without Windows builds finishing because Alpine is good :joy:

PR #498, opened 2023-12-09
Set GOARM=7 explicitly for arm32v7 images

### LINKED ISSUE #494: Something is not right with `GOARM`
`GOARM` value is correct for [`linux/arm/v5` variant](https://hub.docker.com/layers/library/golang/1.21.4/images/sha256-bc69875a9f3af085bd1ea11a44283b46746ef026301552e3c85de7140ad04e90?context=explore):

```
$ docker rmi golang; docker run -q --rm --platform=linux/arm/v5 golang go env GOARM
[…]
5
```

However, it is not correct for [`linux/arm/v7`](https://hub.docker.com/layers/library/golang/1.21.4/images/sha256-fbd184e3e10f5a3e30e01fe8aefb05d7950e0e9a59a6fd0995691b4c239760f3?context=explore):

```
$ docker rmi golang; docker run -q --rm --platform=linux/arm/v7 golang go env GOARM
[…]
6
```

tianon:
This fixes the disparity between the "armv6" download of Go from upstream and the image being v7 otherwise (and thus the default value of `GOARM` being `6` instead of `7`, which is unexpected for users of the image).

This is set in such a way that environment variables (or `go env -w`) will override it, but without requiring us to recompile Go (which I'm not convinced gives us much gain -- happy to reconsider if someone comes up with a good way to benchmark the compiler).

Closes #494

### DIFF
--- a/1.21/alpine3.18/Dockerfile
+++ b/1.21/alpine3.18/Dockerfile
@@ -112,6 +112,15 @@ RUN set -eux; \
 			/usr/local/go/src/cmd/dist/dist \
 			"$GOCACHE" \
 		; \
+	elif [ "$arch" = 'armv7' ]; then \
+		[ -s /usr/local/go/go.env ]; \
+		before="$(go env GOARM)"; [ "$before" != '7' ]; \
+		{ \
+			echo; \
+			echo '# https://github.com/docker-library/golang/issues/494'; \
+			echo 'GOARM=7'; \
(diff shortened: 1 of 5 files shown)


tianon:
Arguably, we should probably add a "verification" block that validates that for a given architecture, all the values in our `env` block are correct from `go env`, but I think that's probably best left as a future enhancement.

yosifkit (approved):

Issue #464, opened 2023-06-10
:tip tag


szabba:
Some people might want to test their code with the latest versions ahead of any official release. It would be nice if that'd be doable by changing just the version tag.

tianon:
Interesting idea -- typically within the official images we're opposed to "master" / "main" / "trunk" / "tip" builds because they typically have a lot of churn, _have_ to build from source, and aren't typically recommended for use by the relevant projects, but in this case, I think `tip` _is_ sufficiently recommended for testing use by the project and the compilation of Go is lightweight enough that this might actually make some sense.

(This _will_ need to make sure it doesn't auto-update too aggressively, but we've used approaches in other places that pin to something like a weekly commit that could be employed here - I'm thinking the day after the [review meetings](https://github.com/golang/go/issues/33502) to maximize the number of features we squeeze into each build, so maybe Thursday mornings?)

I guess that's the long way of saying I'm not opposed, but I'd like to gauge a bit more interest, so hopefully other interested users will see this issue and give you a :+1: reaction. :smile:

AlekSi:
Shameless plug: https://github.com/AlekSi/golang-tip

tianon:
One result of https://github.com/docker-library/golang/pull/500#issuecomment-1863578601 is that this is a little bit harder to implement (both because of the changes here but also the upstream impact of https://github.com/golang/go/issues/54265 now making it progressively harder each release to keep up using only distro packages).

In the end, it probably just means the approach would be different (not just slotting into our current "build from source" code, but a wholly separate "build from source" codepath, likely a separate template because once 1.22 is released and 1.20 is EOL, we won't be building from source at all anymore).

AlekSi:
It would be nice to know how often `tip` is rebuild from the image overview on Docker Hub

tianon:
Good callout, thanks!  It's updated weekly, but we should definitely write that down somewhere easily discoverable.

tianon:
PR up at https://github.com/docker-library/docs/pull/2554

AlekSi:
Thank you!

PR #505, opened 2024-01-16
Use `COPY --link` to share Go layer across OS variants / updates

tianon:
This makes the Go layer reproducible such that re-pulling it can be avoided when switching between OS variants or OS updates.

```console
$ docker buildx build --builder foo --pull --build-arg BUILDKIT_SYNTAX=docker/dockerfile:1 1.21/alpine3.18 --quiet --output type=oci | tar -tv | grep a3881458cc9d5f6798ed03914b3fb322f3620ca53deca3b2c4460676912d16f5
-r--r--r-- 0/0        67061931 1969-12-31 16:00 blobs/sha256/a3881458cc9d5f6798ed03914b3fb322f3620ca53deca3b2c4460676912d16f5
$ docker buildx build --builder foo --pull --build-arg BUILDKIT_SYNTAX=docker/dockerfile:1 1.21/alpine3.19 --quiet --output type=oci | tar -tv | grep a3881458cc9d5f6798ed03914b3fb322f3620ca53deca3b2c4460676912d16f5
-r--r--r-- 0/0        67061931 1969-12-31 16:00 blobs/sha256/a3881458cc9d5f6798ed03914b3fb322f3620ca53deca3b2c4460676912d16f5
$ docker buildx build --builder foo --pull --build-arg BUILDKIT_SYNTAX=docker/dockerfile:1 1.21/bookworm --quiet --output type=oci | tar -tv | grep a3881458cc9d5f6798ed03914b3fb322f3620ca53deca3b2c4460676912d16f5
-r--r--r-- 0/0        67061931 1969-12-31 16:00 blobs/sha256/a3881458cc9d5f6798ed03914b3fb322f3620ca53deca3b2c4460676912d16f5
$ docker buildx build --builder foo --pull --build-arg BUILDKIT_SYNTAX=docker/dockerfile:1 1.21/bullseye --quiet --output type=oci | tar -tv | grep a3881458cc9d5f6798ed03914b3fb322f3620ca53deca3b2c4460676912d16f5
-r--r--r-- 0/0        67061931 1969-12-31 16:00 blobs/sha256/a3881458cc9d5f6798ed03914b3fb322f3620ca53deca3b2c4460676912d16f5
```

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -14,23 +14,9 @@
 		is_alpine and sometimes_build_from_source
 -}}
 {{ if is_alpine then ( -}}
-FROM alpine:{{ alpine_version }}
-
-RUN apk add --no-cache ca-certificates
+FROM alpine:{{ alpine_version }} AS build
 {{ ) else ( -}}
-FROM buildpack-deps:{{ env.variant }}-scm
-
-# install cgo-related dependencies
-RUN set -eux; \
-	apt-get update; \
-	apt-get install -y --no-install-recommends \
-		g++ \
-		gcc \
-		libc6-dev \
-		make \
-		pkg-config \
-	; \
-	rm -rf /var/lib/apt/lists/*
+FROM buildpack-deps:{{ env.variant }}-scm AS build
 {{ ) end -}}
 
 ENV PATH /usr/local/go/bin:$PATH
@@ -74,7 +60,12 @@ ENV GOLANG_VERSION {{ .version }}
 -}}
 RUN set -eux; \
 {{ if is_alpine then ( -}}
-	apk add --no-cache --virtual .fetch-deps gnupg; \
+	apk add --no-cache --virtual .fetch-deps \
+		ca-certificates \
+		gnupg \
+# busybox's "tar" doesn't handle directory mtime correctly, so our SOURCE_DATE_EPOCH lookup doesn't work (the mtime of "/usr/local/go" always ends up being the extraction timestamp)
+		tar \
+	; \
 	arch="$(apk --print-arch)"; \
 {{ ) else ( -}}
 	arch="$(dpkg --print-architecture)"; arch="${arch##*-}"; \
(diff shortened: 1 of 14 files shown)


Dockerfile-linux.template
```diff
@@ -202,6 +201,13 @@ RUN set -eux; \
 			/usr/local/go/src/cmd/dist/dist \
 			"$GOCACHE" \
 		; \
+		\
+# clamp timestamps for reproducibility (allows "COPY --link" to be more clever/useful)
+		date="$(date -d "@$SOURCE_DATE_EPOCH" '+%Y%m%d%H%M.%S')"; \
+		touch -t "$date" /usr/local/.go-date-stamp; \
+		find /usr/local/go -depth -newer /usr/local/.go-date-stamp -exec touch -ht "$date" '{}' +; \
+		rm /usr/local/.go-date-stamp; \
+		touch -t "$date" /usr/local; \
```
LaurentGoderre, line 210: Im wonder if this should be a shared jq template or is it too early for that?

tianon: Too early for that IMO :sweat_smile:

Dockerfile-linux.template
```diff
@@ -166,6 +163,8 @@ RUN set -eux; \
 			apt-get install -y --no-install-recommends -t "$VERSION_CODENAME-backports" golang-go; \
 		); \
 {{ ) end -}}
+# TODO consider CGO_ENABLED=0 instead of installing gcc (probably doesn't matter much as this whole block is 1.20-only and I *think* newer versions already *do* default to cgo-disabled for builds, but maybe that's just for the dist/release builds)
```
tianon, line 166: aha, yes: https://github.com/golang/go/blob/b44f6378233ada888f0dc79e0ac56def4673d9ed/src/cmd/dist/build.go#L1312

however, this was only added in https://github.com/golang/go/commit/da0c375c571037ec5ea6f8ef2be8f07593b40eb6, so only since Go 1.21
(so this is only something to keep in mind when we revisit #464)

tianon: Adjusted the comment to be better

yosifkit (approved):

PR #518, opened 2024-05-22
Alpine 3.20 has been released

techknowlogick:


yosifkit:
Closing in favor of https://github.com/docker-library/golang/pull/519 which has all of the changes.

techknowlogick:
Thanks @yosifkit, that one is indeed the better one to get in :)

Issue #392, opened 2021-09-16
Propose a Centos Based docker image


henri9813:
Could you propose a centos based docker image ?

It could be helpful to test some types of scripts on CentOS directly thanks to a golang:1.17-centos docker image for example

What version of Go are you using (go version)?
every version

What operating system and processor architecture are you using (go env)?
I use Centos :-)

yosifkit:
Duplicate of #160; response is unchanged: https://github.com/docker-library/golang/issues/387#issuecomment-908741129

Issue #387, opened 2021-08-29
[Feauter] Oracle Linux 7/8 support


denismakogon:
Golang on Oracle Linux deserves to exist.
Pros:
- opens a whole world of various flavors of RHEL, Fedora
Cons:
- I'm surprised that no one brought this via issue.

So, what would be nice to have:
- OEL/RHEL 7 and 8
- amd64, arm64 builds

Existing challenges: OEL 7 uses `yum`, OEL 8 uses `microdnf`. The same applies to RHEL/Fedora.

yosifkit:
Basically the same response as https://github.com/docker-library/golang/pull/160#issuecomment-298066350 or https://github.com/docker-library/php/pull/379#issuecomment-282450199. And I'd say this doubly applies for the golang image since it doesn't have to be the runtime environment and you can `COPY` the statically compiled binary to any desired OS image.

tianon:
Here's a simple example to illustrate:

```dockerfile
FROM oraclelinux:8-slim
ENV PATH /usr/local/go/bin:$PATH
COPY --from=golang:1.17 /usr/local/go /usr/local/go/
RUN go version
```

```console
Sending build context to Docker daemon  2.048kB
Step 1/4 : FROM oraclelinux:8-slim
8-slim: Pulling from library/oraclelinux
Digest: sha256:6a1721169ccb64a2a8977112143dd127a7e3a9665fda8774dc2459dd286f87fc
Status: Image is up to date for oraclelinux:8-slim
 ---> 70b244a7aca1
Step 2/4 : ENV PATH /usr/local/go/bin:$PATH
 ---> Running in cc1075ca4452
Removing intermediate container cc1075ca4452
 ---> 9d152600240e
Step 3/4 : COPY --from=golang:1.17 /usr/local/go /usr/local/go/
1.17: Pulling from library/golang
Digest: sha256:4f86bdd4a3118f93a4b84346458d2b88c6e8326ccc81b01d8274daf93fc0b09f
Status: Image is up to date for golang:1.17
 ---> 55fd322d995c
Step 4/4 : RUN go version
 ---> Running in 4a6622f3f41a
go version go1.17 linux/amd64
Removing intermediate container 4a6622f3f41a
 ---> a4c6df3c226a
Successfully built a4c6df3c226a
```

Issue #172, opened 2017-07-24
Support Ubuntu release?


kevinburke:
Hi, we're running Ubuntu in production, and were wondering if it be possible to add/release one?

yosifkit:
Duplicate https://github.com/docker-library/golang/pull/139, https://github.com/docker-library/golang/pull/160

PR #527, opened 2024-07-13
Add support for Rocky Linux 9.3

vborodaykevych:
Add support for Rocky Linux 9.3

- Updated Dockerfile-linux.template to include Rocky Linux 9.3 as a build option.
- Modified apply-templates.sh to handle Rocky Linux 9.3.
- Updated versions.sh to include Rocky Linux 9.3 in the supported versions.
- Modified generate-stackbrew-library.sh to handle Rocky Linux variants and aliases.

This change adds the necessary configurations and scripts to support building Golang images on Rocky Linux 9.3.


tianon:
Sorry, but we have no intention to support a larger matrix of base images than we currently do (see also #139, #160, #172, #387, #392).

You may be interested to note that `/usr/local/go` in our images is already a mostly pristine copy of upstream's release artifacts, so this is trivial for downstream users to replicate on _any_ distro (especially now that they're actually fully static, and thus support Alpine Linux and friends too):

```dockerfile
FROM rockylinux:XXX
ENV PATH /usr/local/go/bin:$PATH
COPY --from=golang:1.22 /usr/local/go /usr/local/go
```

(There might be more bits of https://github.com/docker-library/golang/blob/539882fb23e90d31854a51a773accf8731cf0c9d/1.22/bookworm/Dockerfile you'd want to copy, especially `ENV` statements, but this is the general gist.)

PR #528, opened 2024-07-22
Add go1.22.5-alpine.18

Lockwarr:
[When using wasmvm < v2.1.0, can't build with latest version in go1.22 line with alpine > v3.18](https://github.com/CosmWasm/wasmvm/issues/523)

This PR aims to create a go1.22.5-alpine3.18 image published to the official golang hub.docker so it can be reused easily.


tianon:
Sorry, our policy is to support only two versions of our base distributions, especially given what I described recently in https://github.com/docker-library/golang/pull/527#issuecomment-2243390732 -- namely, the Go installation can be happily copied into other base environments and work as-is (even between Debian and Alpine).

PR #531, opened 2024-08-15
Add "tip" version 

### LINKED ISSUE #464: :tip tag

tianon:
This is pinned to the latest commit merged upstream before midnight (UTC) on Monday of the current week (https://github.com/docker-library/golang/issues/464#issuecomment-1587758290, https://github.com/docker-library/faq#can-i-use-a-bot-to-make-my-image-update-prs).

Closes #464

~~(Includes/based on #530)~~

### DIFF
--- a/Dockerfile-linux.template
+++ b/Dockerfile-linux.template
@@ -13,7 +13,15 @@ FROM buildpack-deps:{{ env.variant }}-scm AS build
 
 ENV PATH /usr/local/go/bin:$PATH
 
+{{ if env.version != "tip" then ( -}}
 ENV GOLANG_VERSION {{ .version }}
+{{ ) else ( -}}
+COPY --from=golang:{{ env.variant }} /usr/local/go /usr/local/goroot-bootstrap
+
+# {{ .version }}: https://github.com/golang/go/tree/{{ .commit.version }}
+ARG GOLANG_COMMIT={{ .commit.version | @sh }}
+ENV GOLANG_COMMIT $GOLANG_COMMIT
+{{ ) end -}}
 
 {{
 	def os_arches:
@@ -54,16 +62,23 @@ RUN set -eux; \
 	now="$(date '+%s')"; \
 {{ if is_alpine then ( -}}
 	apk add --no-cache --virtual .fetch-deps \
+{{ if env.version != "tip" then ( -}}
 		ca-certificates \
 		gnupg \
 # busybox's "tar" doesn't handle directory mtime correctly, so our SOURCE_DATE_EPOCH lookup doesn't work (the mtime of "/usr/local/go" always ends up being the extraction timestamp)
 		tar \
+{{ ) else ( -}}
+		bash \
+		git \
+{{ ) end -}}
 	; \
 	arch="$(apk --print-arch)"; \
 {{ ) else ( -}}
 	arch="$(dpkg --print-architecture)"; arch="${arch##*-}"; \
 {{ ) end -}}
+{{ if env.version != "tip" then ( -}}
 	url=; \
+{{ ) else "" end -}}
 	case "$arch" in \
 {{
 	[
@@ -78,8 +93,12 @@ RUN set -eux; \
 		| (
 -}}
 		{{ $osArch | @sh }}) \
+{{ if env.version != "tip" then ( -}}
 			url={{ .url | @sh }}; \
 			sha256={{ .sha256 | @sh }}; \
+{{ ) else ( -}}
+			export {{ .env | to_entries | sort_by(.key) | map(.key + "=" + (.value | @sh)) | join(" ") }}; \
+{{ ) end -}}
 			;; \
 {{
 		)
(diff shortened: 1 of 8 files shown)


Dockerfile-linux.template
```diff
@@ -107,14 +127,57 @@ RUN set -eux; \
 	\
 # save the timestamp from the tarball so we can restore it for reproducibility, if necessary (see below)
 	SOURCE_DATE_EPOCH="$(stat -c '%Y' /usr/local/go)"; \
+{{ ) else ( -}}
+# before we get too far, let's validate that our "bootstrap" Go works
+	export GOROOT_BOOTSTRAP=/usr/local/goroot-bootstrap; \
+# TODO figure out why QEMU's user-mode emulation with Go 1.23.0 causes "can't start telemetry child process: fork/exec /usr/local/go/bin/go: invalid argument" (it works if we run it a second time, but then later "go build" fails to fork/exec the compile process 🙃)
```
tianon, line 133: because of this "fun" with Go 1.23 + QEMU, I haven't done much testing of this on architectures other than amd64 and i386, so that's a notable blind spot that's probably worth more testing before we consider merging

generate-stackbrew-library.sh
```diff
@@ -111,8 +111,6 @@ for version; do
 
 		fullVersion="$(jq -r '.[env.version].version' versions.json)"
 
-		[[ "$fullVersion" == *.*[^0-9]* ]] || fullVersion+='.0'
```
tianon, line 114: This change is technically valid even without this PR, but it's currently harmless (however with it in this PR, we end up with tags like `tip-YYYYMMDD.0`, which is undesirable).

versions.sh
```diff
@@ -159,7 +242,7 @@ for version in "${versions[@]}"; do
 				"3.19",
 				empty
 			| "alpine" + .),
-			if .arches | has("windows-amd64") and .["windows-amd64"].url then
+			if .arches | has("windows-amd64") and .["windows-amd64"].url then # TODO consider windows + tip
```
tianon, line 245: It's technically _very_ doable for us to implement Windows support, but this was enough that I didn't want to get too much deeper and decided to leave Windows as a TODO for now instead.

tianon:
```diff
$ diff -u <(bashbrew cat golang) <(bashbrew cat <(./generate-stackbrew-library.sh))
--- /dev/fd/63	2024-08-15 16:35:29.584348497 -0700
+++ /dev/fd/62	2024-08-15 16:35:29.584348497 -0700
@@ -2,6 +2,27 @@
 GitRepo: https://github.com/docker-library/golang.git
 Builder: buildkit
 
+Tags: tip-20240810-bookworm, tip-bookworm
+SharedTags: tip-20240810, tip
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: 819b7657d7783ea8b838ad2497af40f785119ccb
+Directory: tip/bookworm
+
+Tags: tip-20240810-bullseye, tip-bullseye
+Architectures: amd64, arm32v5, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
+GitCommit: 819b7657d7783ea8b838ad2497af40f785119ccb
+Directory: tip/bullseye
+
+Tags: tip-20240810-alpine3.20, tip-alpine3.20, tip-20240810-alpine, tip-alpine
+Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, riscv64, s390x
+GitCommit: 819b7657d7783ea8b838ad2497af40f785119ccb
+Directory: tip/alpine3.20
+
+Tags: tip-20240810-alpine3.19, tip-alpine3.19
+Architectures: amd64, arm32v6, arm32v7, arm64v8, i386, ppc64le, s390x
+GitCommit: 819b7657d7783ea8b838ad2497af40f785119ccb
+Directory: tip/alpine3.19
+
 Tags: 1.23.0-bookworm, 1.23-bookworm, 1-bookworm, bookworm
 SharedTags: 1.23.0, 1.23, 1, latest
 Architectures: amd64, arm32v7, arm64v8, i386, mips64le, ppc64le, s390x
```

Dockerfile-linux.template
tianon: This should be fixed now (as of last Oct/1.23.3, actually)! https://github.com/golang/go/issues/68976 :eyes:

tianon: Confirmed, tested successfully:

```console
+ go version
go version devel go1.24-37f27fb Fri Jan 31 08:50:43 2025 -0800 linux/arm64
```

tianon:
With that last push, this should be fully good for review / consideration now. :eyes:

tianon:
Windows CI failures: https://github.com/docker-library/official-images/pull/18435#issuecomment-2654277193

> The problem is that the `windows-2025` runner is too old to run the `mcr.microsoft.com/windows/servercore:ltsc2025` image released this week (https://github.com/actions/runner-images/issues/11582).

yosifkit (approved):
Seems fine.

tianon:
Heh, oops.  This:

https://github.com/docker-library/golang/blob/2a49384d2db5770fb7d909a59c3b6fccec8a35e3/generate-stackbrew-library.sh#L140-L145

Needs to take into account this:

https://github.com/docker-library/golang/blob/2a49384d2db5770fb7d909a59c3b6fccec8a35e3/generate-stackbrew-library.sh#L135-L136

So that we also exclude the _parent_ image's unsupported arches in the case of `golang:tip`.

tianon:
(ie, the "we get lucky" bit there finally flipped to false :joy:)

PR #544, opened 2024-12-13
Simplify and update `verify-templating.yml`

tianon:
This makes it print out a diff when there is an error (instead of just a list of files that are wrong), which will make the error more obvious.  I also added `workflow_dispatch` so we can run it manually.

(If this is unobjectionable, I'll apply it to all our repositories. :+1:)


yosifkit (approved):

tianon:
https://github.com/docker-library/redmine/actions/runs/12324335569/job/34401701901 :+1:

PR #546, opened 2025-01-18
add windows server 2025 support

gdams:
I left 1809 as I believe it's still in extended support


tianon:
Nice, thanks!

> I left 1809 as I believe it's still in extended support

Yeah, I had the same question in https://github.com/docker-library/official-images/pull/18143#issuecomment-2578488987 and came to the same conclusion (ltsc2019 / 1809 is in extended support until 01/09/2029 which is kind of nuts and I hope GitHub drops support for it sooner than that so we can make a good argument for doing so too :grimacing:).

tianon (approved):

PR #549, opened 2025-02-11
Bump to go 1.24.0 release

yosifkit:
Drop end of life `1.22`


techknowlogick:
@yosifkit I think you'll need to bump aliases in `generate-stackbrew-library.sh` too

tianon (approved):

tianon:
Windows failures look like GHA platform failures of some kind (Windows Server 2025 failing to run containers, perhaps because it's Patch Tuesday? :grimacing:)

tianon:
> Windows failures look like GHA platform failures of some kind (Windows Server 2025 failing to run containers, perhaps because it's Patch Tuesday? 😬)

https://github.com/docker-library/official-images/pull/18435#issuecomment-2654277193:

> The problem is that the `windows-2025` runner is too old to run the `mcr.microsoft.com/windows/servercore:ltsc2025` image released this week (https://github.com/actions/runner-images/issues/11582). It should be fine for our builds since we pin to specific Windows images and can delay moving the 2025 pin until the GitHub runners are ready. This will mean that the Windows 2025 images will not get their "Patch Tuesday" rebuilds until the runner images are updated.

PR #553, opened 2025-02-14
Update Git for Windows to 2.48.1

tianon:
It's been a while since we updated this, we probably should (and maybe more often; perhaps automated). :see_no_evil:

### DIFF
--- a/1.23/windows/windowsservercore-1809/Dockerfile
+++ b/1.23/windows/windowsservercore-1809/Dockerfile
@@ -10,13 +10,14 @@ FROM mcr.microsoft.com/windows/servercore:1809
 SHELL ["powershell", "-Command", "$ErrorActionPreference = 'Stop'; $ProgressPreference = 'SilentlyContinue';"]
 
 # install MinGit (especially for "go get")
-# https://blogs.msdn.microsoft.com/visualstudioalm/2016/09/03/whats-new-in-git-for-windows-2-10/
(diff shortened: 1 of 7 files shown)


tianon:
Of course, quite notably, I'm not comfortable merging this right now thanks to https://github.com/docker-library/official-images/pull/18435#issuecomment-2654277193:

> The problem is that the `windows-2025` runner is too old to run the `mcr.microsoft.com/windows/servercore:ltsc2025` image released this week (https://github.com/actions/runner-images/issues/11582).

:upside_down_face: :sob:

tianon:
> perhaps automated

If we use our current automation, this will be updated without being tested (because testing Windows is a blind-spot in our current automation until it pushes the commit and GitHub does a test).

yosifkit:
Close/open to retrigger GitHub Actions now that https://github.com/actions/runner-images/issues/11582 is resolved

yosifkit:
> > perhaps automated
> 
> If we use our current automation, this will be updated without being tested (because testing Windows is a blind-spot in our current automation until it pushes the commit and GitHub does a test).

Yeah, until we include Windows tests in our automated `update.sh` job we should not do it there. Maybe automated PRs to have GHA run the test builds. Perhaps that would just be the `git` updates and not replace the auto-committed `go` version bumps of the `update.sh` job.

yosifkit (approved):

PR #555, opened 2025-02-14
Run upstream tests on images in CI here

tianon:
This runs upstream's `dist test` which is roughly `go test std cmd`.

### DIFF
--- a/.test/config.sh
+++ b/.test/config.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+# https://github.com/docker-library/official-images/blob/3bc6a70175d4e1da2080b86415e6f3c8eb2c6af3/test/config.sh
+
+imageTests[golang]+='
+	go-dist-test
+'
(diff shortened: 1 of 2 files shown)


tianon:
Ouch, I knew they'd be rough, but 10-11 minutes is even heavier than I was picturing.

tianon:
On the upside, total "wall clock" for https://github.com/docker-library/golang/actions/runs/13339366508 (current master) is ~23m 56s and https://github.com/docker-library/golang/actions/runs/13338265228 (this PR) is ~30m 36s.

yosifkit:
I agree to adding tests to ensure that the go binaries work as expected in the container environment but the comparison to the most recent run on master is not quite accurate since it took so long because of the Windows 2025 failures (`cancelled last week in 22m 59s`). Looking at runs before that like https://github.com/docker-library/golang/actions/runs/13297417083 and the complete wall clock is ~14 minutes but greatly slowed by pulling windows images 😢. It's complicated. 🫤

I think the biggest extra slowdown is the serial testing of `1.24` and then `tip`, though I am unsure if running multiple `dist test` at the same time will be significantly faster since it is CPU heavy. But, running the DOI tests in parallel is not something we can change in this repo.

yosifkit (approved):

PR #557, opened 2025-02-25
Slightly different `date` value to get updates Monday

tianon:
I intended for this to update every Monday (with "the last week's changes", including the weekend), but `last monday` doesn't match my intent until *Tuesday*, so this fixes that boundary condition by using `last sunday` instead, one second to midnight, and using date math to add a second.

I also found that even with `--utc`, I had to explicitly specify `UTC` in my string or it would be off by an hour.


tianon:
(this doesn't change `versions.json`, as expected)

yosifkit (approved):

Issue #561, opened 2025-06-11
Support go1.25rc1


lizthegrey:
https://groups.google.com/g/golang-announce/c/dlbl6lgfVX0/m/aD6Br_YFBgAJ?pli=1

```
Hello gophers,

We have just released go1.25rc1, a release candidate version of Go 1.25.
It is cut from release-branch.go1.25 at the revision tagged go1.25rc1.

Please try your production load tests and unit tests with the new version.
Your help testing these pre-release versions is invaluable.

Report any problems using the issue tracker:
https://go.dev/issue/new

If you have Go installed already, an easy way to try go1.25rc1
is by using the go command:

$ go install [golang.org/dl/go1.25rc1@latest](http://golang.org/dl/go1.25rc1@latest)
$ go1.25rc1 download
You can download binary and source distributions from the usual place:
https://go.dev/dl/#go1.25rc1

To find out what has changed in Go 1.25, read the draft release notes:
https://tip.golang.org/doc/go1.25

Cheers,
Cherry and Carlos for the Go team
```

PR #562, opened 2025-06-11
Add 1.25rc1

yosifkit:
Closes https://github.com/docker-library/golang/issues/561

### DIFF
--- a/1.25-rc/alpine3.21/Dockerfile
+++ b/1.25-rc/alpine3.21/Dockerfile
@@ -1,0 +1,128 @@
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "apply-templates.sh"
+#
+# PLEASE DO NOT EDIT IT DIRECTLY.
+#
+
+FROM alpine:3.21 AS build
+
+ENV PATH /usr/local/go/bin:$PATH
+
+ENV GOLANG_VERSION 1.25rc1
+
+RUN set -eux; \
+	now="$(date '+%s')"; \
+	apk add --no-cache --virtual .fetch-deps \
+		ca-certificates \
+		gnupg \
+# busybox's "tar" doesn't handle directory mtime correctly, so our SOURCE_DATE_EPOCH lookup doesn't work (the mtime of "/usr/local/go" always ends up being the extraction timestamp)
+		tar \
+	; \
+	arch="$(apk --print-arch)"; \
+	url=; \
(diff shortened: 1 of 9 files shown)


yosifkit:
Chasing the rabbit hole of the error:
```console
testing golang:1.25rc1-nanoserver-ltsc2022
	'override-cmd' [1/2]...passed
go: open C:\Windows\TEMP: Access is denied.
	'golang-hello-world' [2/2]...failed
```
It seems that the new version of go tries to access the `$TMP` directory, which **should** be readable and writable by the running user.

`nanoserver` images have these values in their environment variables
```console

TEMP=C:\Windows\TEMP
TMP=C:\Windows\TEMP
USERNAME=ContainerUser
```

😭 Every `mcr.microsoft.com/windows/nanoserver` image, from the [first](https://oci.dag.dev/?image=mcr.microsoft.com%2Fwindows%2Fnanoserver:10.0.17763.316) `1809` (aka `ltsc2019`) through `ltsc2025` has `TMP` and `TEMP` set to the system temporary directory, but the default user (`ContainerUser`) doesn't have permission read its contents (or write to it).

Earlier `nanoserver` images (like `1803`) have the correct setting based on the user (e.g `C:\Users\ContainerUser\AppData\Local\Temp` for the default user) and it auto adjusts if you run as `ContainerAdministrator` which is how the `windows/servercore` images work.

Adding `-e TMP=C:\Users\ContainerUser\AppData\Local\Temp -w C:\Users\ContainerUser\AppData\Local\Temp` is enough to get `go` working again (i.e., set the temp env var and make sure the directory exists).

So, we have a few options that I see:
- disable the nanoserver builds on 1.25-rc
- set `TMP` and `TEMP` via `ENV` in the Dockerfile (or maybe `setx` like `PATH`?)
  - what value works for any user?
- set `USER ContainerAdministrator`
  - 👎less secure
- give ContainerUser read/write permissions to `C:\Windows\temp`
  - this also seems less secure
- convince windows to fix the nanoserver images to auto set `TMP` and `TEMP` like they did in `1803`

---

related:
- https://github.com/PowerShell/PowerShell-Docker/issues/188
- https://github.com/microsoft/artifacts-credprovider/issues/448

yosifkit:
For now, I've added a commit to disable nanoserver images for `1.25rc1` (which means we'll have to figure something out by rc2).

Issue #570, opened 2025-08-21
`cannot find 'ld'` on Trixie


AlekSi:
arm64 M1 MacBook Pro, macOS 15.6 (24G84), Docker Desktop 4.44.3 (202357), Docker Engine 28.3.2.

Installing `golangci-lint` works successfully on Bookworm:
```
$ docker run --rm golang:1.25-bookworm go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest
```

It fails with Trixie:
```
$ docker run --rm golang:1.25-trixie go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest
go: downloading github.com/golangci/golangci-lint/v2 v2.4.0
[...]
go: downloading github.com/matttproud/golang_protobuf_extensions v1.0.1
# github.com/golangci/golangci-lint/v2/cmd/golangci-lint
/usr/local/go/pkg/tool/linux_arm64/link: running gcc failed: exit status 1
[...]

PR #571, opened 2025-08-21
Explicitly install `gold` on `arm64`

tianon:
- Fixes https://github.com/docker-library/golang/issues/570
- https://github.com/golang/go/issues/22040

> WARNING: gold is being removed from binutils, and is deprecated upstream.


tianon:
I've tested `1.25/trixie` successfully with the `golangci-lint` example from the linked issue -- doing much more than that is heavy because I'm not on `arm64` and the emulation is very slow. :smile:

yosifkit (approved):

PR #582, opened 2026-06-10
Add Alpine 3.24 and drop 3.22

jnoordsij:
Modelled after #577.

This PR updates Alpine to the latest stable version: 3.24.

See also https://alpinelinux.org/posts/Alpine-3.24.0-released.html.


tianon (approved):
Nice, thanks!

